PR #55, opened 2019-04-22
Add documentation with modular steps management

lfoussat:


doc/object-creation.md
```diff
@@ -0,0 +1,53 @@
+# Admin object's management - create an object
+
+## Usage
+> Elements of the app are managed threw objects in *Admin*.
```
bde-maze: Threw from throw, like a ball. I think you want to use through, like through the day

doc/child-object-creation.md
```diff
@@ -0,0 +1,33 @@
...
+
+#### 3. Set up the new child:
+* Enter its name in the input "Add a child name"
+* Select your step object in the select input
+* Click on "ADD"
+
+> Your step is then related to its parent. You can see it in the *Children* section of the parent's object. There, you can now: 
+> * Delete the child from its parent (the actual object of your child will not be deleted).
+> * Reorder it in the children's list, by dragging it to the place you want.
+> * Go to update its settings by clicking on the eye of its reference (it will redirect you to the actual object of your step).
```
bde-maze: may be worth it to precise which settings. Also you mention the steps, but I think this page of documentation is not about the onboarding steps but only about the creation of a child, could be confusing. I would for something like so. 
Update its *original* settings by clicking on the eye icon of its reference (redirection to object edit page of the child).

doc/object-creation.md
```diff
@@ -0,0 +1,53 @@
...
+
+> Objects of the Admin are first created and defined:
+> * By their **title**,
+> * By their **type**.
+
+> Then it can be configured threw:
```
bde-maze: here as well

lfoussat: true though - corrected!

doc/object-creation.md
lfoussat: true though - corrected

doc/child-object-creation.md
lfoussat: updated

lfoussat:
first version of this doc is ready for merge, I think. I'm waiting for your review

PR #143, opened 2019-05-22
update of quest 1 subjects in french and english

Frenchris:
Update of sujects list with quest 1 and adding of all the subject of quest1 (bash code)


subjects/cl-camp9.en.md
```diff
@@ -0,0 +1,32 @@
+# cl-camp8
```
xpetit: `cl-camp8` but the name of the file is `cl-camp9`

PR #180, opened 2019-06-17
Ztail readme

augusto-mantilla:


Frenchris:
put those modfications in another bracnh

PR #197, opened 2019-06-20
change function to adapt to Go language

augusto-mantilla:


Frenchris:
did another pull request taking this changes into account

PR #271, opened 2019-10-01
Create specs-500-students-school.md

bde-maze:
This document describes the minimum specs for a 500 students school (server and wifi)


kigiri:
no update for a while not sure that we need that, reopen if needed @bde-maze

PR #273, opened 2019-10-01
lastrune program for the exam

augusto-mantilla:


Frenchris:
I grouped the subject from lastrune on the nruneprg public to match the same sequence than the PRs on the all repository. I am closing this branch without merging it to avoid conflicts and ensure a proper track keeping of the files.

PR #290, opened 2019-10-03
foreach readme for the exams

LEEDASILVA:


Frenchris:
This program differs too much from its initial exercise. Let's see with augusto if it can be adapted otherwise. (same comment than on all repo # 458)

PR #331, opened 2019-11-18
Fillit project readme

OGordoo:


subjects/fillit.en.md
```diff
@@ -0,0 +1,109 @@
...
+All files referring to the project, must be in the same folder.
+
+The program must:
+
+- Be written in **GO**
+- Be compiled to be runned
+- Assemble all of the **Tetrominos** in order to create the smallest square possible
+- Show an error if the input doesn't correspond to a classic **Tetromino** piece
+- Not allow the rotation of the **Tetrominos**
+- Organize the **Tetrominos** by their apparition order in the file
```
xpetit: I suggest we remove this instruction, to allow different implementations and to keep only one objective here : make the smallest square with the tetrominoes.

subjects/fillit.en.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Instructions
+
+All files referring to the project, must be in the same folder.
+
+The program must:
+
+- Be written in **GO**
+- Be compiled to be runned
+- Assemble all of the **Tetrominos** in order to create the smallest square possible
```
xpetit: In the wikipedia article, the plural of tetromino seems to be more often tetrominoes.
I also suggest not to treat the word tetromino as a special word, by that I mean not putting it in bold or capital letters.

subjects/fillit.en.md
```diff
@@ -0,0 +1,109 @@
...
+- Show an error if the input doesn't correspond to a classic **Tetromino** piece
+- Not allow the rotation of the **Tetrominos**
+- Organize the **Tetrominos** by their apparition order in the file
+- Identify each **Tetromino** in the solution, by assigning different charactes to different **Tetrominos**
```
xpetit: Typo: charactes -> characters
Also you can be more specific here by using the word "letters"

subjects/fillit.en.md
```diff
@@ -0,0 +1,109 @@
...
+- Have at least 2 **Tetrominos** in the text file
```
xpetit: This list starts with "The program must:"
So I'm reading "The program must have at least 2 **Tetrominos** in the text file" which is inaccurate.

subjects/fillit.en.md
```diff
@@ -0,0 +1,109 @@
...
+```console
+ABB.
+ABB.
+A...
+A...
+```
+
+
+
+- Among all the possible candidates for the smallest square, we only accpet the ones where **Tetrominos** is placed on their most upper-left position. In the example below, the program would choose option a).
```
xpetit: Like my previous comment, I suggest that we remove this type of constraint, this will allow some students to try new algorithms.

subjects/fillit.en.md
OGordoo: What we mean is that text file should have at least two pieces (Tetrominos).

subjects/fillit.en.md
OGordoo: Shouldn't it be "runes"? Because they can choose numbers or special characters.
I was going towards that thought, that's why I wrote "characters".

xpetit: oh nice idea, this way we could test their code with a file containing tetrominoes made of emojis 🤭

subjects/fillit.en.md
xpetit: so this line should be :
- Expect at least 2 tetrominoes in the text file
so that the student will read : "The program must expect at least 2 tetrominoes in the text file"

subjects/fillit.en.md
```diff
@@ -0,0 +1,94 @@
+## fiil it
```
xpetit: typo

subjects/fillit.en.md
```diff
@@ -0,0 +1,94 @@
...
+### Instructions
+
+All files referring to the project, must be in the same folder.
+
+The program must :
+
+- Be written in Go
+- Compile successfully
+- Assemble all of the Tetrominoes in order to create the smallest square possible
+- Make the smallest square with the Tetrominoes
```
xpetit: Redundant with the previous line

PR #340, opened 2019-11-21
test -> sortparams

zhantolin:


kigiri:
Thanks a lot @zhantolin good catch

PR #341, opened 2019-11-24
Clarify rot14

atlekbai:
On exam not all students faced rot13 before.

### DIFF
--- a/subjects/rot14.en.md
+++ b/subjects/rot14.en.md
@@ -14,7 +14,7 @@ func Rot14(str string) string {
 
 ### Usage
 
-Here is a possible program to test your function :
+Here is a possible [program](TODO-LINK) to test your function :	
 
 ```go
 package main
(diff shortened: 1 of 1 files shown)


subjects/rot14.en.md
```diff
@@ -2,9 +2,7 @@
 
 ### Instructions
 
-Write a function `rot14` that returns the `string` within the parameter transformed into a `rot14 string`.
-
--   There is the need to know what `rot13` stands for.
```
kigiri: I believe that this subject was supposed to not be that clear, you can easly google rot13 and try to figure it out.
Also the broken program links are currently placeholders the links will be replaced, unrelated to the issue.

atlekbai: Students cant google for rot13 during the exam. They were asking what rot13 is, cuz they didnt do it before.

PR #350, opened 2019-11-26
changing overflow cases

MSilva95:


subjects/doop.en.md
```diff
@@ -18,9 +18,9 @@ In case of an invalid operator the programs prints `0`.
 
 In case of an invalid number of arguments the program prints nothing.
 
-In case of overflow the program prints `Overflow`.
+In case the result gives overflow the program prints `Overflow`.
 
-The program has to handle the modulo and division operations by 0 as shown on the output examples below.
+The program has to handle the module and division operations by 0 as shown on the output examples below.
```
xpetit: Why change "modulo" into "module" ?

subjects/doop.en.md
```diff
@@ -42,7 +42,7 @@ No Modulo by 0$
 student@ubuntu:~/piscine-go/test$ ./doop 9223372036854775807 + 1
 Overflow
 student@ubuntu:~/piscine-go/test$ ./doop -9223372036854775809 "*" 3
-Overflow
+0
```
xpetit: Why ? -9223372036854775809 * 3 appear as an overflow to me but maybe I'm wrong

MSilva95: -9223372036854775809 * 3 this is already an overflow so i changed the calculation to subtract instead of multiply

xpetit: Ok, I'm noticing another error : `-9223372036854775808` is the smallest `int64`, not `-9223372036854775809` (invalid)

MSilva95: its to show that if one of the arguments is out of range by the atoi it will print "0" because its an overflow

xpetit: In that case even `-9223372036854775809 + 0` will fail, because Atoi will trigger an error since `-9223372036854775809` is an invalid number (like `123abc` for example).
It is not an integer overflow.

MSilva95: yes, and in both cases the output will be 0
that case it's just to show that if you put a number that is too big(invalid) the program prints "0"

PR #361, opened 2019-11-30
2 is not power of 2 number

azakost:
x^2 = 2
what should be x then?

ALSO testing program says that number 8388608 should be true
x^2 = 8388608
what should be x then?

### DIFF
--- a/subjects/ispowerof2.en.md
+++ b/subjects/ispowerof2.en.md
@@ -15,7 +15,7 @@ This program must print `true` if the given number is a power of 2, otherwise it
 ```console
 student@ubuntu:~/ispowerof2$ go build
 student@ubuntu:~/ispowerof2$ ./ispowerof2 2 | cat -e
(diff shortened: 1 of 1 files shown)


subjects/ispowerof2.en.md
```diff
@@ -15,7 +15,7 @@ This program must print `true` if the given number is a power of 2, otherwise it
 ```console
 student@ubuntu:~/ispowerof2$ go build
 student@ubuntu:~/ispowerof2$ ./ispowerof2 2 | cat -e
-true$
```
LEEDASILVA: correct me if i'm wrong, but 2^x = 2
that means that x = 1 so it is power of 2, right?
http://www.tsm-resources.com/alists/pow2.html

azakost: I see. I've just translated task differently. Thanks now i understand =)

PR #362, opened 2019-11-30
Could be two outcome

azakost:
By following all clues eventually i found two suspect who could be killer
- Dartey Henv
- Hellen Maher

1. Both are not in People list (so we cannot say what gender they are)
2. Both driving Blue Honda with plates starting L377 and ending 9
3. Both more than 6' tall

I make a personal assumption that Helen is female, so killer is Dartey Henv. But it is according to my assumption, not clues I have.

### DIFF
--- a/subjects/now-get-to-work.en.md
+++ b/subjects/now-get-to-work.en.md
@@ -19,3 +19,4 @@ student@ubuntu:~/piscine-go/test$
 ### Hint
 
 "You could combine `head` and `tail`s..."
+


Frenchris:
CLUE: Footage from an ATM security camera is blurry but shows that the perpetrator is a tall **male**, at least 6'. 

One of the clues mentions that the suspect is a male. 
Hellen is a female name. No assumption to be made.

PR #446, opened 2020-01-21
standarD instead of standarT

IrvinIrvin:
All D changed to T


LEEDASILVA:
hello it's standard not standart

PR #452, opened 2020-02-01
Update nauuo.en.md

Danonika:
Input is always positive.
0 is not positive.


Frenchris:
Nauuo subject to be fully reviewed and tested for clarification. Will check in next exam test.

PR #483, opened 2020-03-02
Forum

OGordoo:


subjects/forum/forum.audit.en.md
```diff
@@ -0,0 +1,143 @@
...
+
+###### Did the server behaved as expected?(did not crashed)
+
+###### Does the server use the right [HTTP method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods)?
+
+###### Has the website runned without crashing at anytime?
+
+###### Are all the pages working? (Absence of 404 page?)
+
+###### Does the project avoid [HTTP status 400](https://kinsta.com/knowledgebase/400-bad-request/#causes)?
```
xpetit: Confusing, what do you mean ?

subjects/forum/forum.audit.en.md
```diff
@@ -0,0 +1,143 @@
...
+
+###### Does the project avoid [HTTP status 500](https://www.restapitutorial.com/httpstatuscodes.html)?
```
xpetit: Same here

subjects/forum/forum.audit.en.md
OGordoo: What I meant by that is to see if in the project  they are handling bad requests (HTTP status 400) and internal server errors (HTTP status 500).

xpetit: It would then be preferable to speak of "handle" and not "avoid", as some Alem students told us, "avoid" can be understood in this way: "make sure it's impossible to get that HTTP status"

OGordoo: ok, so i will change it to handle, thanks

subjects/forum/forum.en.md
```diff
@@ -0,0 +1,101 @@
...
+  - Compatibility/Dependency.
+  - Creating images.
+- SQLite language.
+  - Manipulation of databases.
+- The basics of encryption.
+
+### Instructions
+
+- You must use **SQLite**.
+- You must use HTML files.
```
kigiri: I find `use HTML` to be not really necessary to specify, it's a web forum so HTML will be served but you don't really use the html yourself, you produce it for the client.
What were you trying to say here ?

subjects/forum/forum.en.md
```diff
@@ -0,0 +1,101 @@
...
+### Objectives
+
+This project consists in creating a web forum that allows :
+
+- communication between users and the community through the creation of posts/comments.
+- associate posts to categories.
+- non-registered users to only see posts/comments.
+- likes and dislikes in posts and comments, you must take in consideration that the number of likes and dislikes must be accounted.
+- searching, inside your forum, for a specific category or post
+  - The search bar must have typing suggestions as you write.
```
kigiri: I don't think type suggestions as you type can be done without JS, would not ask for it

subjects/forum/forum.en.md
```diff
@@ -0,0 +1,101 @@
...
+
+This project will help you learn about:
+
+- Client utilities.
+- The basics of web :
+  - HTML
+  - HTTP
+  - Sessions and cookies
+- Using and [setting up Docker](https://docs.docker.com/get-started/) :
+  - Services and dependencies.
```
kigiri: Do we still have to mention services here since it was removed ?

subjects/forum/forum.en.md
OGordoo: I was trying to say that the students will have to use HTML, but as you said is kinda pointless. Maybe take it off?

subjects/forum/forum.en.md
OGordoo: no, I thought that I had removed it. Miss that one

subjects/forum/forum.en.md
LEEDASILVA: It's possible using the datalist from html and go templates, but i prefer to remove it and do it for the next forum? because in JS its better

PR #491, opened 2020-03-05
searchPosition

OGordoo:


Frenchris:
Close along with PR # 717 from all repo

PR #517, opened 2020-04-04
fixed bash syntax for ARG

azamatamirzhan:


azamatamirzhan:
Right syntax to ARG:
ARG=("arg1 arg2 ...");

PR #521, opened 2020-04-10
adding time-limit on tetris-optimizer

LEEDASILVA:


subjects/tetris-optimizer/tetris-optimizer.audit.en.md
```diff
@@ -48,15 +48,15 @@
 
 ##### Try good example 02 (https://public.01-edu.org/subjects/tetris-optimizer/goodexample02).
 
-###### Does the result contain 4 empty spaces (4 '.')?
+###### Does the result contain 4 empty spaces (4 '.') and a time limit lesser or equal to 0,018s?
```
xpetit, line 51: That's too short, for some reason the computer might freeze right at the moment of running the program and the student and it might create unnecessary discussion. Also I would prefer to have a generic high limit, like 1, 5 or 10 minutes, regardless of the map tested. Or we can even put a rule in the good practices or somewhere else to say that any computing program must deliver in less than an amount of time at least of one minute.
Also think that this time is dependent of the computer single-core performance. So putting very precise timing like this (or in lem-in.audit.en.md) is a mistake in my opinion. Maybe a future school will run with mini PC and a low-consumption CPU. Also the auditor can wait 1 min, can't he ?

LEEDASILVA: ok ok will add a standard time of 5 minutes on the good practices page

LEEDASILVA: i think that 5 minutes is ok for computing programs. What do you think?

xpetit: ok that will teach them patience 😂

PR #531, opened 2020-05-13
Add audit question for several categories

anr-m:
The forum should have an option to associate several categories to a post, as per the task description. It should be checked in the audit.


LEEDASILVA:
thanks anr-m that question has already been added to the audit

PR #544, opened 2020-06-02
Update more-or-less.en.md

atlekbai:
Clarification of subject


kigiri:
I don't find this clearer, since in javascript you can not pass a reference on primitive values, it's a given that you must return the value to add one to it.

If the student is confused about that I find it preferable to make him realise that this is the only way to do it in js.

I don't want them to write code like:
```js
const add = n => {
 n++
 return n
}
```
and think that it's the same value.

PR #546, opened 2020-06-02
Update change.en.md

atlekbai:
Add clarification, fix links

### DIFF
--- a/subjects/change.en.md
+++ b/subjects/change.en.md
@@ -3,20 +3,21 @@
 ### Instructions
 
 Create 2 functions:
-- `get`: a function that takes a key and return the corresponding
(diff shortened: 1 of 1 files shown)


subjects/change.en.md
```diff
@@ -3,20 +3,21 @@
 ### Instructions
 
 Create 2 functions:
-- `get`: a function that takes a key and return the corresponding
+- `get`: a function that takes a key and returns the corresponding
```
kigiri, line 6: I don't believe `return` ever takes an `s` before `the`.
same for update, please don't fix what's not broken haha.

Other changes were made with batch fixes.

PR #550, opened 2020-06-03
Update nasa.en.md

atlekbai:
Add clarification to subject. It's unclear whether we should return array or string.


kigiri:
I think `split` would make me think it's an array, i'm using `separated` instead, thanks for the suggestion

PR #551, opened 2020-06-03
Update triangle.en.md

atlekbai:
Fix output of the program


kigiri:
because of conflicts it was easier to just redo the change, thanks for the PR @atlekbai

PR #554, opened 2020-06-04
Update find-expression.en.md

sakenism:


subjects/find-expression.en.md
```diff
@@ -25,6 +25,6 @@ It will look something like this `1 *2 *2 +4`
 > all code provided will be added to your solution and doesn't need to be submited.
 
 ```js
-const add4 = '+4'
-const mul2 = '*2'
+const add4 = ' +4'
```
kigiri, line 28: While the expected results are separated by space, this does not change the value of the given constants.
You may redeclare them if you want to add a space, it's not needed for the exercise.

PR #580, opened 2020-06-11
Testing in that way

azamatamirzhan:
Test checks array structure.
The task description doesn't gives the right way to do it well without "oh snap!")


kigiri:
Thanks for the correction

PR #582, opened 2020-06-12
beginEnd case correction

azamatamirzhan:
'hired kimchi' has 'hi' in the beginnig and in the end

### DIFF
--- a/js/tests/its-a-match_test.js
+++ b/js/tests/its-a-match_test.js
@@ -21,7 +21,7 @@ t(() => !end.test('Hi'))
 t(() => end.test('I love sushi'))
 
 t(() => beginEnd.test('hi'))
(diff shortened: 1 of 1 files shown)


OGordoo:
HI @azamatamirzhan, `beginEnd` should be true if the same instance of 'hi' is simultaneously in the beginning and in the end

kigiri:
> HI @azamatamirzhan, `beginEnd.test(...)` should be true if the same instance of 'hi' is simultaneously in the beginning and in the end

maybe we can try to make this clearer in the subject @OGordoo ?

azamatamirzhan:
'hired kimchi' has 'hi' in the end, and in the beginning too.
Written program must return true - am I right?

OGordoo:
@kigiri Okay, I'm going to make it clearer on the subject

PR #595, opened 2020-06-16
fix typi

kuzikov:
53 weeks is 371  day - more than there is in a year.


kuzikov:
nvm, i misunderstood this task.

PR #600, opened 2020-06-17
Fix code in flow

atlekbai:
It took me 10 min to realize why I can't execute the code xD


kigiri:
wow even with the change I had too look at it 3 times before seeing the difference haha

PR #602, opened 2020-06-17
Sortable Raid

OGordoo:


subjects/sortable/README.en.md
```diff
@@ -0,0 +1,34 @@
...
+
+You are a villain and your dream is to end with those annoying, yoga pants,
+weird masks wearing superheroes. You never understood why are some of them
+considered superheroes just because they are rich. Others annoy you with their
+philosophical speeches. And of course that something tragic has had to happen
+to them for the people to feel sorry for them. Anyway, we've found confidential
+information about those superheroes.
+
+Your task for the moment is to build a web page in order to organize all the
+data from those smartypants. This information can be found [here](https://ogordoo.github.io/flatHeroAPI/superheroAPI.json).
```
kigiri, line 13: the data should be hosted in the subject directory named `data.json`

subjects/sortable/README.en.md
```diff
@@ -0,0 +1,34 @@
...
+
+- All the data must be contained in a table
+  - Images should be shown too, as images and not as a string.
+- The information must be displayed in multiple pages displaying from 10, 20,
+  50, 100 or all results
+- It must be possible to sort all columns of the table (either alphabetically
+  or numerically)2
+  - If you sort more than once, it should display the results in
```
kigiri, line 21: I think this could be more specific:

> First click will order on this column by ascending value. \
> Consecutive clicks will toggle between ascending and descending.

We could also specify that it should initially be ordered by id, asc.

Also it would be nice to ask them to persist the selection choices using `localStorage`

subjects/sortable/README.en.md
```diff
@@ -0,0 +1,34 @@
...
+
+Your task for the moment is to build a web page in order to organize all the
+data from those smartypants. This information can be found [here](https://ogordoo.github.io/flatHeroAPI/superheroAPI.json).
+
+- All the data must be contained in a table
+  - Images should be shown too, as images and not as a string.
+- The information must be displayed in multiple pages displaying from 10, 20,
+  50, 100 or all results
+- It must be possible to sort all columns of the table (either alphabetically
+  or numerically)2
```
kigiri, line 20: you have a wierd 2 in here.

I would ask them to handle units, like in values that have `19 kg`,
we should ask that in ascending order:
the right order should be `['2 kg', '19 kg']` and not `['19 kg', '2 kg']`

Order by value, not simplistic string compare.

PR #631, opened 2020-06-23
content: issue fix

LEEDASILVA:


subjects/printstr/README.md
```diff
@@ -31,6 +31,6 @@ And its output :
 ```console
 student@ubuntu:~/[[ROOT]]/test$ go build
 student@ubuntu:~/[[ROOT]]/test$ ./test | cat -e
-Hello World!%
+Hello World!$
```
xpetit, line 34: Now the test compare student's function to this one :
```go
func printStr(s string) {
	fmt.Print(s)
}
```
So I think it was ok as it is, and then subjects doesn't talk about new line, so `%` is expected

LEEDASILVA: ok @xpetit, this was one of the issues that students posted. will correct it, thanks

PR #641, opened 2020-06-25
Update README.md, add link css-tricks#throttle

atlekbai:
Add link about throttle

### DIFF
--- a/subjects/throttle/README.md
+++ b/subjects/throttle/README.md
@@ -11,3 +11,4 @@ Create two functions that will work like `_.throttle` from lodash
 ### Notions
 
 - [lodash.com/docs/4.17.15#throttle](https://lodash.com/docs/4.17.15#throttle)
+- [https://css-tricks.com/debouncing-throttling-explained-examples/#throttle](https://css-tricks.com/debouncing-throttling-explained-examples/#throttle)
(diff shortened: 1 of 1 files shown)


subjects/throttle/README.md
```diff
@@ -11,3 +11,5 @@ Create two functions that will work like `_.throttle` from lodash
 ### Notions
 
 - [lodash.com/docs/4.17.15#throttle](https://lodash.com/docs/4.17.15#throttle)
+- [https://css-tricks.com/debouncing-throttling-explained-examples/#throttle](https://css-tricks.com/debouncing-throttling-explained-examples/#throttle)
+- [https://stackoverflow.com/questions/24079736/confused-about-the-maxwait-option-for-lodashs-debounce-method](https://stackoverflow.com/questions/24079736/confused-about-the-maxwait-option-for-lodashs-debounce-method)
```
LEEDASILVA, line 15: I'm not sure about this. We agreed that the link put in notions are specific links, like "devdocs.io" or "lodash". I think that if students search for it they will be able to find it, what do you think?

atlekbai: https://github.com/01-edu/public/tree/master/subjects/sweet-curry has link in Notions to stackoverflow. I think it will be useful.

PR #654, opened 2020-07-06
adding spaces in the output

MSilva95:


go/tests/prog/grouping_prog/main.go
```diff
@@ -45,17 +45,16 @@ func brackets(regexp, text string) {
 		runes = runes[1 : len(runes)-1]
 		result := simpleSearch(runes, text)
 		for i, s := range result {
-			if !unicode.Is(unicode.Hex_Digit, rune(s[len(s)-1])) {
-				s = s[:len(s)-1]
+			if !unicode.Is(unicode.Hex_Digit, rune(s[len(s)-1])) && !unicode.IsLetter(rune(s[len(s)-1])) {
+				s = s[0 : len(s)-1]
```
xpetit, line 49: Those two lines are the same :

```go
s = s[0 : len(s)-1]
s = s[:len(s)-1]
```

go/tests/prog/grouping_prog/main.go
```diff
@@ -45,17 +45,16 @@ func brackets(regexp, text string) {
 		runes = runes[1 : len(runes)-1]
 		result := simpleSearch(runes, text)
 		for i, s := range result {
-			if !unicode.Is(unicode.Hex_Digit, rune(s[len(s)-1])) {
-				s = s[:len(s)-1]
+			if !unicode.Is(unicode.Hex_Digit, rune(s[len(s)-1])) && !unicode.IsLetter(rune(s[len(s)-1])) {
```
xpetit, line 48: if
```go
!unicode.IsLetter(rune(s[len(s)-1]))
```
is true then
```go
!unicode.Is(unicode.Hex_Digit, rune(s[len(s)-1]))
```
is always true

go/tests/prog/range_prog/main.go
```diff
@@ -22,8 +20,14 @@ func main() {
...
-	for _, i := range lib.IntRange(a, b) {
-		fmt.Print(i)
+	fmt.Print(a)
+	for b != a {
+		if a < b {
+			a++
+		} else {
+			a--
+		}
+		fmt.Print(" ", a)
```
xpetit, line 30: If you tested it and are 100%-ish sure that it works, you can push/merge master directly

go/tests/prog/grouping_prog/main.go
MSilva95: I removed one of the ifs that were repeated and tested the exercise several times according to the readme and the tests and everything is working now

go/tests/prog/range_prog/main.go
MSilva95: ok i will merge it then, thank you

go/tests/prog/grouping_prog/main.go
xpetit: Yes it is working but what I meant is that those conditions are certainly similar :

```go
if !unicode.Is(unicode.Hex_Digit, rune(s[len(s)-1])) && !unicode.IsLetter(rune(s[len(s)-1])) {
if !unicode.IsLetter(rune(s[len(s)-1])) {
```

but the latter is simpler, don't you think ?

MSilva95: ohhh, yes you are right, i tested it only with the first condition and it didn't work, but if we do it with the second one only it works fine, my bad... 
do you think we need to change this?

xpetit: Yes it is important to simplify code ^^ it makes maintenance and reading easier. You can modify the line, add/commit/push on master !

PR #667, opened 2020-08-12
mini-framework project

OGordoo:


subjects/mini-framework/README.md
```diff
@@ -0,0 +1,103 @@
...
+  <div class="nameSubm">
+    <input type="text" placeholder="Insert Name" />
+    <input type="submit" placeholder="Submit" />
+  </div>
+</html>
+```
+
+The html above can be written as:
+
+```json
```
kigiri, line 35: json can't have comments

OGordoo: It's corrected, thanks

subjects/mini-framework/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+Now that you have already used a framework of your choice, you must now implement some features on a framework of your own. That's right, you are going to create a framework.
+
+Be aware that a framework is different from a library. When you call a method from a library, you are in control. But with a framework, the control is inverted: the framework calls you.
+
+### Objectives
+
+Your framework should implement:
+
+- Abstracting DOM
```
kigiri, line 11: should be:
` - Abstracting the DOM`

subjects/mini-framework/README.md
```diff
@@ -0,0 +1,105 @@
...
+- Easier Routing
```
kigiri, line 12: Not very important but I don't like `Easier` as a way to describe the work needed to be done for the routing... but I'm not sure about what would be better so haha !

Maybe: `Routing System` ?

OGordoo: Yeah honestly, routing system seems better ahah

subjects/mini-framework/audit/README.md
```diff
@@ -0,0 +1,41 @@
+#### Functional
+
+###### Does the project contain documentation? (If so, you will need it)
```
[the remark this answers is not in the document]
OGordoo: Yes it has to be mandatory so that the user has some guide to do the tests

subjects/mini-framework/README.md
```diff
@@ -0,0 +1,105 @@
...
+- Abstracting the DOM
+  Routing System
+- State Management
+- Event Handling
+
+### Instructions
+
+You must create documentation for your framework, so that users (auditers) are able to understand and know how to use your framework without experiencing any awkwardness. By documentation we mean, the explaining of how does the framework works and how to work with it, for example: how to create a div, how to add an event to a button, etc.
+
+Your framework will be tested by using it, like you previously have used one, in the social network project. So the user has to be presented to a folder structure that when executed `npm start` at the root of the folder, it runs the app. The user testing your framework will have to implement some simple code in order to test the features described bellow.
```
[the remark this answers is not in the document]
OGordoo: Yeah agreed, i have removed it

subjects/mini-framework/routing/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+The example consists in switching between a **'Home Page'** to a **'Active Page'** by clicking on a **'Click Me'** hyperlink.
+
+You will need to two or three files depending if you include the script on the html files or not. All files should be in the same folder.
+
+index.html:
+
+```html
+<!DOCTYPE html>
+<html class="no-js">
```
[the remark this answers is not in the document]
OGordoo: in the html we can ignore, but in the other elements is to see if their framework is placing the classes in the elements

subjects/mini-framework/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+There are a lot of ways to achieve this. Above are just some examples, what matters is that the DOM must respond to certain actions of the user.
+
+You have to take into account the events, children and attributes of each element of the DOM.
+
+---
+
+#### Routing System
+
+Routing is the process of selecting a path for traffic in a network. And, as you may have already realized, routing in JS can sometimes be a little bit confusing, so what you will need to do, is to implement a new routing system that is more intuitive and simpler.
```
kigiri, line 84: in this case, routing is not a a path for traffic in a network though.
It's more, a synchronization of the state of the app with the url, not sure how to describe it well but:

`/user/23` could be the page of the user `id:23` and `/user/23/edit` could be the same page but with a modal open where you can edit the data inside.

For example in todoMVC, you can have filters like `active` reflected in the `URL`, even if 0 network requests are made.

subjects/mini-framework/routing/README.md
```diff
@@ -0,0 +1,62 @@
...
+    <script src="script.js" async defer></script>
+  </body>
+</html>
+```
+
+script.js:
+
+```js
+const readTextFile = (file, allText = '') => {
+  let rawFile = new XMLHttpRequest()
```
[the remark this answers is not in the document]
kigiri: still not using `fetch` ? why ?

subjects/mini-framework/README.md
```diff
@@ -0,0 +1,112 @@
...
+- State Management
+- Event Handling
+
+You will also need to make a [todoMVC](http://todomvc.com/) project using your framework.
+
+### Instructions
+
+You must create documentation for your framework, so that users (auditers) are able to understand and know how to use your framework without experiencing any awkwardness. By documentation we mean, the explaining of how does the framework works and how to work with it, for example: how to create a div, how to add an event to a button, etc.
+
+Your framework will be tested by using it, like you previously have used one, in the social network project. So the user has to be presented to a folder structure that allows him to run the app from the root of that folder structure. The user testing your framework will have to implement some simple code in order to test the features described bellow.
```
kigiri, line 22: I think documentation should have it's own big title so it's not overlooked by students.
We should make a few points like:
Must be in markdown
Include code example
a complete reference that cover all the framework API
explain the reason why things are done like this.

PR #670, opened 2020-08-31
graphql project

LEEDASILVA:
**Beside** the **subject** and **audit** I added a page that gives information about the **graphQL schema** that is given to them.
This page is for the student to use so that he/she has a perspective of which tables and fields are permitted to use.


subjects/graphql/README.md
```diff
@@ -0,0 +1,185 @@
+## GraphQL
+
+### Objectives
+
+The objective of this project is to learn graphql query language by creating profiles. It will be provided,\
```
kigiri, line 5: graphQL = graph query language, so repeating query language after is a bit wierd, maybe putting it after would sound less repeated:

> The objective of this project is to learn the query language [graphQL](https://graphql.org/) by creating your profile page. It will be provided,\

subjects/graphql/README.md
```diff
@@ -0,0 +1,185 @@
...
+- Basic user identification
+- XP amount
+- level
+- grades
+- skills
+- statistics
+
+### Instructions
+
+> Here is a small introduction and advantages of graphQL, if you want to read more about graphQL you can visit there [site](https://graphql.org/)
```
kigiri, line 20: I think it's better to leave the link at the first mention on graphQL and not after.

subjects/graphql/README.md
```diff
@@ -0,0 +1,185 @@
+## GraphQL
+
+### Objectives
+
+The objective of this project is to learn graphql query language by creating profiles. It will be provided,\
+by the platform, a graphql endpoint that is connected to the database. So you can query this endpoint to obtain the information you desire.\
```
kigiri, line 6: graphql -> graphQL, let's keep this consistent

subjects/graphql/README.md
```diff
@@ -0,0 +1,185 @@
...
+**Note** that for security reasons some tables are private and some are public, you will only be provided with certain content.
+
+Your profile must have at least:
+
+- Basic user identification
+- XP amount
```
kigiri, line 12: not sure I want to force students to show their level / xp, maybe we can let them be more free about what they want to display on this profile ?

what do you think ?

subjects/graphql/README.md
```diff
@@ -0,0 +1,185 @@
...
+- level
+- grades
+- skills
+- statistics
+
+### Instructions
+
+> Here is a small introduction and advantages of graphQL, if you want to read more about graphQL you can visit there [site](https://graphql.org/)
+
+GraphQL is a query language for any API endpoint and runtime. The syntax lets you specify what data you want to receive from that API.\
```
kigiri, line 22: I don't think we have to explain what graphQL is, they can look it up.

subjects/graphql/README.md
```diff
@@ -0,0 +1,185 @@
...
+You must have in mind that this language is used for API endpoints and not database. Unlike SQL,\
+graphql query does not go to your database directly but to your graphql API endpoint which can connect to a database or any other data source.
+
+You are already familiar with REST, since REST is a robust methodology of creating APIs and not elastic/scalable.\
```
kigiri, line 26: saying that rest is not scalable and graphQL is a bit bullshit talk, both can be scaled very well or very badly.

REST is very easy to cache, and this allow for very strong scalability if done properly.

`elastic` ? what does that mean ?

GraphQL strong point is an expressive query language in the client, this has a huge bunch of draw backs too, harder to limit / restrict and handle permissions.

Having a query language directly in the client then allow:
- retrieve only what you want
- batch queries in a single request
- move a lot of backend code to the client, as such helping with being sure query code match the actual client need and keeping the codebase clean.

I think you have 3 big wins:
- Front end teams don't need to ask back end teams to get what they want, so less friction for big teams
- Less data over the network, this is important for market with unreliable connections (3g phone, india, africa...)
- Speed of development, it's easier and faster to make queries in graphQL so you can move faster in your development.


Main trade of are:
- more code shipped too the front end side, heavier client applications, more complexity.
- much more ressource intensive for the servers, having to parse the query, and no caching make you do a lot more work
- as it is an abstraction that make it easier to query, you can easly do very expensive queries without them looking very complex, this is a hidden cost of the "fast & easy" queries,

subjects/graphql/README.md
```diff
@@ -0,0 +1,185 @@
...
+It can be at the same time painful, because it requires individual creation of each API, example: `v1/user/item/{id}`, `v1/post/item/{id}` and so on.
+
+The main feature of graphql compared to REST is that it lets you ask for specific information. And even better then that is the nesting feature.
```
kigiri, line 29: But overall, this is not a blog about is graphQL good / bad and when / why use it.
I'm not sure it's the role of this document to talk about those.

I think we just want them to know about it so they can make their own mind about what to think of it.

subjects/graphql/README.md
```diff
@@ -0,0 +1,185 @@
...
+Any other information you desire to display is welcome and will be noted
+
+Beside this information, you will have to create a search engine which returns a selection of students profiles, based on the students:
+
+- Name
+- XP
+- level
+- skills
+
+For instance you can search for a student by their `githubLogin` or filter the students by the amount of `XP` or which `level`/`skill` they have.
```
kigiri, line 56: I plan on renaming `githubLogin` to just `login` too

subjects/graphql/README.md
```diff
@@ -0,0 +1,185 @@
...
+    user {
+      id
+      githubLogin
+    }
+  }
+}
+
+```
+
+For this example we ask for the results `id` and `user`s that are associated to the `result`, requesting the users `name`s and `id`s.\
```
kigiri, line 179: I think instead of talking that much about `graphQL` we should also talk about the UI a bit, maybe ask for them to generate statistic graph using `SVG` and maybe to a bit about hosting ?
where / how should they host it ?

Since xavier was against rolling our own githubpage, maybe we can ask them to publish it on github-page ? or netlify or some other free static hosting service of their choice.

subjects/graphql/README.md
```diff
@@ -0,0 +1,185 @@
...
+- **Basic user identification**, for example **githubLogin**
+- **XP** amount
+- **level**
+- **grades**
+- **skills**
+- **statistics**
+
+Any other information you desire to display is welcome and will be noted
+
+Beside this information, you will have to create a search engine which returns a selection of students profiles, based on the students:
```
kigiri, line 49: A search engine !?
How do we handle that on our side ?
I think it's more of a UI problem than a query problem, I don't know... I don't think graphQL is a good fit for a search type query system.

You can definitly do it but, still not sure of the overall value.

subjects/graphql/README.md
```diff
@@ -0,0 +1,185 @@
...
+}
+
+```
+
+For this example we ask for the results `id` and `user`s that are associated to the `result`, requesting the users `name`s and `id`s.\
+
+This project will help you learn about:
+
+- [Graphql](https://graphql.org/) language
+- Custom search operations (`include`/`exclude`/`fuzzy`)
```
kigiri, line 184: `%ilike%` is not fuzzy search.
Fuzzy is a backend problem they can't do it without direct access to the database to add function and extend hasura to get fuzzy search features, which we do not provide.
(and if we did they wouldn't learn about it, just use it)
- https://hasura.io/blog/full-text-search-with-hasura-graphql-api-postgres/

subjects/graphql/index.html
```diff
@@ -0,0 +1,119 @@
...
+
+</style>
+</head>
+<body>
+  <script type="module" src="./lib/schema.js"></script>
+  <div class="docExplorerWrap">
+      <div class="doc-title"><h3 class="title">Schema Documentation</h3></div>
+      <div class="doc-explore-content">
+        <label class="search-box">
+          <img src="https://cdn1.iconfinder.com/data/icons/thin-ui-1/100/Noun_Project_100Icon_1px_grid_thin_ic_cookie-512.png">
```
kigiri, line 108: please use inline SVG instead of images for icons

subjects/graphql/lib/schema.js
```diff
@@ -0,0 +1,163 @@
+const backImg =
+  'https://cdn1.iconfinder.com/data/icons/thin-ui-1/100/Noun_Project_100Icon_1px_grid_thin_ic_arrow_left_simple-512.png'
```
kigiri, line 2: the `<` character was not good enough ?

subjects/graphql/index.html
```diff
@@ -0,0 +1,119 @@
...
+<html>
+<head>
+  <title>Graphql Doc</title>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+<style>
+
+:root {
+  --color: hsl(0, 0%, 65%);
+}
```
kigiri, line 11: what is all this ?
Did you recode your own [GraphiQL](https://github.com/graphql/graphiql) ?

subjects/graphql/lib/schema.json
```diff
@@ -0,0 +1,136 @@
+{
```
kigiri, line 1: why not query the schema from graphQL instead of having it hardcoded ?
This is a good way to make sure it's out of date soon

kigiri (changes requested):
Not sure I understand everything, You did a lot of work but I would need explaination about why it was needed what purpose does it all serve ?

subjects/graphql/README.md
LEEDASILVA: ok noted, will fix it

subjects/graphql/README.md
LEEDASILVA: yes its a good idea, but the main reason i added this was for the audit questions, that way the questions can be more functional (Is the amount of XP for this student displayed on the profile?).

I can say that they must display on their profile at least 4 sections of content at their choice??

subjects/graphql/README.md
LEEDASILVA: ok, on it

subjects/graphql/README.md
LEEDASILVA: yha, i did try to say something like that but it seams that i failed. so i will just delete everything thats explaining graphql and the comparing to REST to graphql. Thank you for the points on graphql

subjects/graphql/README.md
LEEDASILVA: ok will change it on the subject

subjects/graphql/index.html
```diff
@@ -1,119 +1,37 @@
...
+
+    <script>
+      const graphQLFetcher = graphQLParams =>
+        fetch(`https://${location.hostname}/api/graphql-engine/v1/graphql`, {
+          method: 'post',
+          headers: { 'Content-Type': 'application/json' },
+          body: JSON.stringify(graphQLParams),
+        })
+          .then(response => response.json())
+          .catch(() => response.text());
```
kigiri, line 30:
```js
   .catch(() => response.text())
```
wat ?

Here `response` is not defined, also you don't do  anything with the error.

I believe it's best not to catch at all and let `graphiQL` handle the error.
> edit: Ok i've check graphiql examples, we should do:
> ```js
> .then(response => response.json().catch(() => response.text()))
> ```

It's only if the `.json()` fail that we fallback to text

kigiri (commented):
All right other than that it's all good for me !

subjects/graphql/index.html
LEEDASILVA: ok ok will correct it, thanks

kigiri (approved):

PR #672, opened 2020-09-09
Bomberman dom

OGordoo:


subjects/ascii-art/justify/README.md
```diff
@@ -13,8 +13,7 @@ You must follow the same [instructions](https://public.01-edu.org/subjects/ascii
 
 This project will help you learn about :
 
-- Client utilities.
-- The Go file system(**fs**) API.
+- The Go file system(**fs**).
```
kigiri, line 16: I think it was better with API, or maybe library or functions ?
Why did you decided to remove it ?
It's not like it's the go filesystem, a language do not provide a file system.

subjects/bomberman-dom/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+   1. The maps should be generated with a logical random structure (without blocking the player). There will be two types of blocks, the ones that can be destroyed and the ones that can not. Tip: the optional project [different maps](https://public.01-edu.org/subjects/make-your-game/different-maps) can be useful for this part.
+   2. The players should be placed in the corners as their starting positions.
+
+3. Power ups (each time a player destroys a block, a random power up may or may not appear):
+
+   1. Bombs: Increases the amount of bombs dropped at a time by 1;
+   2. Flames: Increases explosion range from the bomb in four directions by 1 block;
+   3. Speed: Increases movement speed;
+   4. Bomb Push: Ability to throw a bomb after it has been placed.
```
kigiri, line 44: Maybe bomb push could be a bonus ?

Ideas for bonuses:
- Solo + CO-OP mode (play against AI)
- more power ups !
- release power ups after death
- more than 4 players
- team mode
- after death interaction (once dead players can use the environment as ghosts or something to try to kill other players, if they succeed they get back alive in their place)

subjects/bomberman-dom/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+1. Players
+
+   1. Nº of players: 2 - 4
+   2. Each player must have 3 lives. Then you are out!!
+
+2. The map should be fixed so that every player sees the whole map.
+
+   1. The maps should be generated with a logical random structure (without blocking the player). There will be two types of blocks, the ones that can be destroyed and the ones that can not. Tip: the optional project [different maps](https://public.01-edu.org/subjects/make-your-game/different-maps) can be useful for this part.
+   2. The players should be placed in the corners as their starting positions.
```
kigiri, line 37: about randomly generated, I think most bomberman maps are only half random.
Walls are fixed but boxes (breakable walls) are put at random.

It is way easier because they don't have to worry about blocking other players since players can break them, just blacklist the starting corners.

kigiri (commented):
looking good !

subjects/ascii-art/justify/README.md
OGordoo: i see an API more like a "software" that has no connection to a language. Maybe library or package is more suitable, what do you think?

kigiri: API is how you interact with it, code is an API for reference `https://nodejs.org/api/fs.html` this is the link of the node API which contains code to interact with the FS, you can look at occurrences of `API` in the page and find quite a lot.

OGordoo: ooooh ok, got it, then i will leave API in there

kigiri (approved):

LEEDASILVA (approved):

PR #674, opened 2020-09-10
Update README.md

Danonika:
Push-swap subject:
*The goal is to sort the stack with the minimum possible number of operations.
Update answer on push-swap example from 8 to 7 operations.


OGordoo:
That's absolutely correct, thank you!

PR #689, opened 2020-12-20
fix: expect wrong result by search  "queen"

devstackq:


LEEDASILVA:
Queensland is a location not an artists <https://groupietrackers.herokuapp.com/api/locations/4>

PR #697, opened 2021-01-21
road_intersection: rust s raid 2

LEEDASILVA:
- subject and audit for rust raid 02
- contains some subjects for rust exercises

### DIFF
--- a/subjects/road_intersection/README.md
+++ b/subjects/road_intersection/README.md
@@ -36,8 +36,8 @@ _______________________          |          ____________________
                        |    ↓    |         |
 ```
 
-There are 4 incoming lanes. Each lane has a set of consecutive unique ids `(ri, ri, ri)` where `(i = 0, 1, ⋯, 11)`, numbered in clockwise starting from the top-most. Each set of ids
-indicates the outgoing direction of a vehicles. Vehicles driving in a given lane will have a given **route** (r0 | r1 .... r11) that indicates the lane destination of that vehicles.
+There are 4 incoming lanes. Each lane has a set of consecutive unique ids `(ri, ri, ri)` where `(i = 0, 1, ⋯, 11)`, numbered clockwise starting from the top-most. Each set of ids
++indicates the outgoing direction of a vehicle. Vehicles driving in a given lane will have a given **route** (r0 | r1 .... r11) that indicates the lane destination of that vehicles.
 
 For this raid you must follow these assumptions:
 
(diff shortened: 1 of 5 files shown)


subjects/road_intersection/README.md
```diff
@@ -7,71 +7,106 @@ Its up to you to decide which library and file system you want do use to create
...
+               |    ↓    |    ↑    |
 ```
 
-There are 4 incoming lanes. Each lane has a set of consecutive unique ids `(ri, ri, ri)` where `(i = 0, 1, ⋯, 11)`, numbered clockwise starting from the top-most. Each set of ids
-indicates the outgoing direction of a vehicle. Vehicles driving in a given lane will have a given **route** (r0 | r1 .... r11) that indicates the lane destination of that vehicles.
+For clarification reasons we will assume that a lane can have three different routes (consider you are in the vehicle position):
+
+- `r0`, turning right
+- `r1`, straight ahead
+- `r2`, turning left
```
augusto-mantilla, line 47: What if we say `r`, for right. `l`, for left and `s` for straight? I think is a bit more clear

subjects/road_intersection/README.md
```diff
@@ -7,71 +7,106 @@ Its up to you to decide which library and file system you want do use to create
...
+Vehicles must obey this rules:
 
-#### **Stop conditions**
+- Autonomous, vehicles driving on a lane with a **given route** must 
+follow the direction of that route, its not possible for the driver to change lanes or route.
 
-As you know vehicles must not collide, but for **debugging and testing** reasons you must show some kind of feedback (logs, terminal display or canvas display its up to you to decide)
-if vehicles collide, so that the users can conclude that the vehicles collided.
-Each lane will have a traffic light system (either imaginary of real), the objective is that each vehicle stops if there is any type of collision with other vehicles. Otherwise the vehicles can proceed in their lane.
+- Each vehicle must have a fixed velocity.
```
augusto-mantilla, line 71: Maybe we can add a distinction for the car so we can see the route that they are suppose to follow. Maybe, a different color for each route. Like red cars turn left. Or something like that. So the reviewer can check that the car is indeed following the correct route.

augusto-mantilla (changes requested):

subjects/road_intersection/README.md
LEEDASILVA: ok ok done

subjects/road_intersection/audit/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+###### Did all vehicles passed the intersection safely?
+
+##### Try to generate two vehicles at the same time, one using the `"Right"` key and the other using the `"Left"` key (to this at least 3 times).
+
+###### Did all vehicles passed the intersection safely?
+
+##### Try to generate two vehicles at the same time, one using the `"Up"` key and the other using the `"Left"` key (to this at least 3 times).
+
+###### Did all vehicles passed the intersection safely?
```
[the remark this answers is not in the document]
augusto-mantilla: Another options is to define that if you can see the shapes of the cars intercepting then that means that there was a collision and the intersection was not passed safely. If we want to make it easier for the students

augusto-mantilla (changes requested):

subjects/road_intersection/audit/README.md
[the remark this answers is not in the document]
LEEDASILVA: sens the raids is becoming a bit big i will go with the second option. Adding the following to the audit
` ###### Did all vehicles passed the intersection safely and without any collision?` and maybe add the definition of collision

subjects/road_intersection/README.md
LEEDASILVA: about the colors i agree. we can let them choose the colors they want, but it was to be different colors for each route

augusto-mantilla (approved):

PR #705, opened 2021-01-29
Post category in Forum

azamatamirzhan:
Posts without category forbiden.


Frenchris:
@LEEDASILVA what do you think of this adding in the forum subject? Does a post has to have at least one category or can a post be uncategorized? This aspect is not clear in the audit questions for me.

LEEDASILVA:
Post can be uncategorized, there is no need to have at least one category for a post. The categories where to help users to filter  posts, and help students on the creation of several ways of doing so. It is also good for testing the algorithm you do for the filters.

PR #710, opened 2021-02-08
Rust subject review quest 02

Frenchris:
1) Issues with subject strings_lirerals.
 a)The subject seems incomplete, the naming of the folder in the repository was wrong (corrected in this PR) 
 b)The Usage part is empty. 

2) Issues with doubtful (how ironic, haha).
a) The subject asks for a function, the expected files are set to recieve a main.rs, is this a mistake from a migration?
b) The subject states that the code has to be "fixed where indicated", however no such indications were mentioned. I may need some assistance for correcting this one without changing the intended purpose.  

3) Issues with the "heap exercises" arrange_it and name_initial: 
- I try to fail name_initials with a working solution asking too much heap and I get an unexpected answer from the server. 5/6 times (it read timeout once)
- I try to pass arrange_it with an almost identical solution than the rust-solutions repo and I got an unexpected answer from the server. 
- I made Xavier aware of those issues see if he can do something as this type of exercise really need the error messages to help the debugging.

4)Ongoing issue with permissions when import are done outside the src.rs or main.rs file, Augusto and Xavier were made aware. Awaiting some news for retesting `borrow_me_the_reference`.


augusto-mantilla (commented):
I corrected the readme for doubtful.  The expected file for `doubtful` should be `doubtful/src/lib.rs`. @LEEDASILVA should take a look at the string_literal exercise

subjects/string_literals/README.md
```diff
@@ -0,0 +1,53 @@
+## string literal
```
LEEDASILVA: for the `string_literal` exercise, the name here was correct the wrong naming was done on the migration. I will avoid the migration and change the name in the rest of the files

LEEDASILVA (commented):
i removed the duplicated subject and added some corrections

subjects/string_literals/README.md
augusto-mantilla: @Frenchris is this pull request finished? I would like to merge this pull request before you make new ones. I don't want it to get messy with a bunch of pull requests open at the same time.

augusto-mantilla (approved):

augusto-mantilla:
I accepted the changes and fix the problems that @Frenchris found. I also reverted some of the changes, where does changes didn't improve the subject.

PR #711, opened 2021-02-10
Rust subject review quest 03

Frenchris:
### There is an error with testing with the platform ( I put the correct solution). 
```console Caused by:
  failed to load source for dependency `rand`

Caused by:
  Unable to update registry `https://github.com/rust-lang/crates.io-index`

Caused by:
  failed to fetch `https://github.com/rust-lang/crates.io-index`

Caused by:
  network failure seems to have happened
  if a proxy or similar is necessary `net.git-fetch-with-cli` may help here
  https://doc.rust-lang.org/cargo/reference/config.html#netgit-fetch-with-cli

Caused by:
  failed to resolve address for github.com: Temporary failure in name resolution; class=Net (12)```

## Strings:
### The test has a naming issue with the function
```console
Issue with the test: char_lenght instead of char_length
```

### issue with the test (tried the right solution)
console ```
error: failed to get `collect` as a dependency of package `collect_test v0.1.0 (/jail/tests/collect_test)` Caused by: failed to load source for dependency `collect` Caused by: Unable to update /jail/solutions/collect Caused by: failed to read `/jail/solutions/collect/Cargo.toml` Caused by: No such file or directory (os error 2)```

##The rest is working. 
### Small edits.
- Did edits to expected function and usage to make them work.
- Added some notions links but not too many.

### DIFF
--- a/subjects/card_deck/README.md
+++ b/subjects/card_deck/README.md
@@ -17,9 +17,9 @@ Represent the cards from a deck:
 
 Define:
 
-- The associated **function** `transpose` for `Rank` and `Suit`:
-  - For `Suit`, `transpose_suit` makes the translation between an integer value(u8) and the suit of a card (1 -> Heart, 2 -> Diamonds, 3 -> Spade, 4 -> Club)
-  - For `Rank`, `transpose_rank` makes the translation between an integer value(u8) and the rank ( 1 -> Ace, 2 -> 2, .., 10 -> 10, 11 -> Jack, 12 -> Queen, 13 -> King)
+- The associated **function** `translate` for `Rank` and `Suit`:
+  - For `Suit`, `translate` makes the translation between an integer value (u8) and the suit of a card (1 -> Heart, 2 -> Diamonds, 3 -> Spade, 4 -> Club)
+  - For `Rank`, `translate` makes the translation between an integer value (u8) and the rank ( 1 -> Ace, 2 -> 2, .., 10 -> 10, 11 -> Jack, 12 -> Queen, 13 -> King)
 - The associated **function** `random` for `Rank` and `Suit` which returns a random `Rank` and `Suit` respectively
 - Finally define the **function** `winner_card` which returns `true` if the card passed as an argument is an Ace of spades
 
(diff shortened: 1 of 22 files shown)


subjects/card_deck/README.md
```diff
@@ -2,46 +2,44 @@
...
+[Crate rand](https://docs.rs/rand/0.5.0/rand/)
 
 ### Expected Functions and Structures
 
 ```rust
 pub fn random() -> Suit {
 }
 
-pub fn translate(value: u8) -> Suit {
+pub fn transpose_suit(value: u8) -> Suit {
```
augusto-mantilla, line 36: Yes, there was a typo. Both should be name `translate`. There is no need to add the prefix _card and _deck because associated functions are how the name says associated to the type. So you will have to call them like this: Card::translate_card which is actually bad practice. Notice the useless repetition of the word card.

subjects/simple_hash/README.md
```diff
@@ -35,16 +35,16 @@ fn main() {
 
     println!(
         "Does the HashMap contains the name Roman? => {}",
-        contain(&hash, "Roman")
+        contain(hash.clone(), "Roman")
```
augusto-mantilla, line 38: There is no need to clone the hash map. Just borrowing it is enough. I'm afraid this is going to enforce a sub-optimal way of writing Rust. There is no need to take ownership of a value if you only want to read the its content. But this is more with @LEEDASILVA who made the exercise

augusto-mantilla (changes requested):

augusto-mantilla:
# About the exercise strings
I corrected the name to `char_length`
# About the exercise card_deck
I corrected the typo in the associated function `translate`

subjects/simple_hash/README.md
LEEDASILVA: hey i did not do this exercise, maybe @MSilva95?

augusto-mantilla: Ah ok, my bad @LEEDASILVA . I went and I change it myself. The solution, test and subject are all up to date now.

augusto-mantilla (approved):

PR #715, opened 2021-02-11
add links to generics and traits

anr-m:
students who don't know about generics or traits may be confused and stuck on quo and rem


augusto-mantilla:
There is no need to use neither generics nor traits to solve this exercise

anr-m:
Sorry, didn't realize that the github version is already updated. There was no hints for the types in the subject, so I had to use generics 😄

augusto-mantilla:
No problem :smile: . I'm closing now

PR #719, opened 2021-02-11
review first quest of blockchain

MSilva95:


xpetit:
We already have 3 different cases for subjects names, please don't add camelCase :D

PR #722, opened 2021-02-14
Rust subject review quest 04

Frenchris:
Quest 4
- profanity_filter timing out with solution from the tests. (probably because of the external crate use of chrono::prelude::Utc

- error_types, (subject not fetching)
Checked the tests, it is singular(error_type) the test needs to be adapted to plural.
Changed the key to plural ( will need to be corrected in a migration)
Changed the folder in public to plural. 

Not understanding the password requirements here: 

**The first name must not be empty and the password must have at least 8 char and a combination of alphanumeric and none alphanumeric ASCII characters**
_None alphanumeric ASCII characters??_ What does that mean?? 
The examples are unclear for me ( the first and the last one)
ex: "asDd123=%" => good
   "asgfD"     => error
   "asdsdf2"   => error
   "sad_#$"    => error

**Once this subject is fixed** I will need to recheck it. 

- Boxing_todo

The error trait should state “failed to parse todo” (needs to be corrected in the solutions of the tests in file error.rs)

The crate serde_json = "1.0" is not imported in the test.
Also there are some issues with the number of arguments of get_todo, and read_todos, parse todos when testing the usage main program.  Need help getting that one fixed. I tried with the solution from the tests. 

- The rest

The rest seems good. Did some typo correcting, formatting and added some resources on the other subjects.


augusto-mantilla:
@Frenchris I believe it's suposed to be **non-alphanumeric**.

LEEDASILVA:
@Frenchris will rename the folders and files that needs renaming (`error_type`).
----
For the `profanity_filter` its strange that it gives timeout will take a look into in
----
Like @augusto-mantilla mentioned it's suposed to be non-alphonumeric, the example gives the student the combination of the password:

ex: "asDd123=%" => good (all alphanumeric non-alphanumeric)
"asgfD" => error (not good because it does not have non-alphanumeric)
"asdsdf2" => error (does not include some kind of ascii `%` for example)
"sad_#$" => error (does not include any digit)
---
The `boxing_todo` seems to have the error trait stating `Failed to parse todo` instated of `Fail to parse toido` i will fix it
for the implementation of `serde`, it is implemented in the test, i do not see the problem.
i will take a look into the subject to see whats the problem with the usage

LEEDASILVA (commented):
@Frenchris, i corrected the subjects and made a PR on `rust-test` repo. Fixed the usage of box_todo because it was using deprecated functions. You can review the subjects and see if everything is correct

augusto-mantilla (approved):

PR #723, opened 2021-02-15
Update the adder call

biblion84:
Added is called with two arguments here : 
adder([9, 24, 7, 11, 3], 10)
But the instruction given tells us that added only take one argument, an array
added is not called elswhere with two arguments
Thus : is this a typo or do the students needs to handle two arguments as well ?


biblion84:
disregard did not see the rest of the instruction

PR #725, opened 2021-02-16
Rust subject review quest 05

Frenchris:
- middle_day

There was no section expected function but I guessed it was because the student has to find out the signature of the function.

I added one section
Expected Function 

With the message: 
For this exercise the signature of the function has to be found out.

- does_it_fit

I do not think that the person who wrote this one realized that the title of this exercise has a sexual innuendo. ( showed it to clem, it s not just me who saw it right away :) )
It needs to be changed to something else. 
Maybe call it 
areas_volumes_fit   or area_volumes_fitter

Whichever is chosen,
Let’s not forget to change it in the title of the readme and the import in the usage program. 

Also I added the module areas_volumes to the subject ( it said will be given).

- Macro_calculator

Error when testing with real solution

Crates update issue
```console
  Updating crates.io index warning: spurious network error (2 tries remaining): failed to resolve address for github.com: Temporary failure in name resolution; class=Net (12) warning: spurious network error (1 tries remaining): failed to resolve address for github.com: Temporary failure in name resolution; class=Net (12) error: failed to get `serde` as a dependency of package `macro_calculator_test v0.1.0 (/jail/tests/macro_calculator_test)` Caused by: failed to load source for dependency `serde` Caused by: Unable to update registry `https://github.com/rust-lang/crates.io-index` Caused by: failed to fetch `https://github.com/rust-lang/crates.io-index` Caused by: network failure seems to have happened if a proxy or similar is necessary `net.git-fetch-with-cli` may help here https://doc.rust-lang.org/cargo/reference/config.html#netgit-fetch-with-cli Caused by: failed to resolve address for github.com: Temporary failure in name resolution; class=Net (12)
```

- Shopping_mall:

I added the module mall to the subject ( it said will be given).
I added the Expected function section
With the message
For this exercise the signature of the function has to be found out.

- Expected_variable:

Issues when testing **locally** with rust-solutions:
I get an infinite loop in the third test of the Usage program, the one that is supposed to display Fail.

Also the formatting of the output of the Usage test is wrong 

Issue with test **on the platform**: 
```console
No test file found for the exercise : expected_variable
```
I believe that is because the test file expected_variable_test is missing the  _test
I am not sure if they may be any other issue inside. I prefer to let the writer solve this one along with the infinite loop i get locally. 

- Mobs:

Inconsistencies between, the files to submits in the attributes and the solutions in the rust tests

This is in the rust test (and it works when submitted)

mobs/src/mob.rs 
mobs/src/mob/boss.rs 
mobs/src/mob/member.rs 

This in the expected files

 mobs/src/mobs.rs 
mobs/src/mobs/    

Ohterwise super cool subjects. Quite challenging.

### DIFF
--- a/subjects/banner/README.md
+++ b/subjects/banner/README.md
@@ -2,17 +2,16 @@
 
 ### Instructions
 
-`Result` is a better version of the `Option` type that describes possible error instead
-of possible absence
+"`Result` is a better version of the `Option` type that describes possible `error` instead of possible `absence`".
 
-Create a structure called `Flag` that as the following elements:
+Create a structure called `Flag` which has the following elements:
 
-- short_hand: String
-- long_hand: String
-- desc: String
+- short_hand: `String`
+- long_hand: `String`
+- desc: `String`
 
-This structure must have associated to it a function called `opt_flag` that initializes the structure.
-Receiving two references strings and returns the structure `Flag`. It should be used like this:
+This structure must have a **function** called `opt_flag` which initializes the structure.
+This **function** receives two strings references and returns a structure `Flag`. Here is an example of its usage:
 
 ```rust
     let d = Flag::opt_flag("diff", "gives the difference between two numbers");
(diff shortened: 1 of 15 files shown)


LEEDASILVA:
will assign @OGordoo since he made this quest

OGordoo:
- does_it_fit

"There are no dirty words, only dirty minds"
Ok, we can turn it into `areas_volumes_fitter`

- macro_calculator

I think that is a network error. I tried to test disconnected from the internet and the same exact error appeared.

- Files Changed

I noticed that there are files changed that don't belong to the Quest 05, but do belong to Quest 04. I don't know what happened there.

xpetit:
@Frenchris @OGordoo In my opinion "does_it_fit" should not be changed because some males have a misplaced mind.

If you google "does it fit" (with safe search disabled) you can even see an Harvard document... So please **guys** take it easy

Frenchris:
> @Frenchris @OGordoo In my opinion "does_it_fit" should not be changed because some males have a misplaced mind.
> 
> If you google "does it fit" (with safe search disabled) you can even see an Harvard document... So please **guys** take it easy
@xpetit @OGordoo haha, now I feel like I may need to go to some anonymous meetings. Ok, let's keep it the original way then :)

Frenchris:
"macro_calculator
I think that is a network error. I tried to test disconnected from the internet and the same exact error appeared."

@OGordoo , yes. I did not have the issue locally, only on the platform. I tried again, it still the same error. Maybe somewhere in the test the import of serde is not well written? I checked the Cargo.toml file of the solutions and it seems ok.

xpetit:
@Frenchris Apple's guys are far worse :
Tell Siri "You're a slut" and Siri answers "I would blush if I could"
Screenshot_2021-02-16_15-19-01

Frenchris:
haha, My favorite is Alexa's answer to the "You're a slut"

augusto-mantilla (approved):

PR #726, opened 2021-02-16
Patch 2

Zulbukharov:
update Expected Functions to public


LEEDASILVA:
Thank you for the correction, but this subject has already been corrected

PR #731, opened 2021-02-19
Update README.md

Frenchris:
little typos fix proposal.


Frenchris:
I pushed directly into a commit.

PR #733, opened 2021-02-21
fix example code in rust/string_literals

anr-m:


LEEDASILVA:
Thank you for the correction on `string_literals` subject

PR #736, opened 2021-02-23
Update README.md

rtabulov:
clarify number of arguments

### DIFF
--- a/subjects/interpolation/README.md
+++ b/subjects/interpolation/README.md
@@ -2,7 +2,7 @@
 
 ### Instructions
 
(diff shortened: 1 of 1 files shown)


LEEDASILVA:
is this really necessary? "that takes an object with 5 properties" this means that the function will take just one argument

rtabulov:
well if it means "take just one argument", then why not just write it? 
it is also a matter of consistency - other tasks say how many arguments they need

but then again, it's your call :D feel free just to close it

kigiri:
Yeah I don't think this one add much, I like subject to be concise if possible, at least in the instructions

PR #737, opened 2021-02-23
get-json task clarifications

rtabulov:
add a few important clarifications into the task get-json description

### DIFF
--- a/subjects/get-json/README.md
+++ b/subjects/get-json/README.md
@@ -5,15 +5,15 @@
 In this exercise, we will focus on building complex async flows with promises.
 
 Create a `getJSON` function that takes 2 parameters:
-- `path`, that will be the url called by your function
-- `params` *optional*, that will be the search parameters appended to your url
+- `path`, string that will be the url called by your function
+- `params` *optional*, object that will be the search parameters appended to your url
 
 `getJSON` must construct a valid url with the `path` and stringified `params`
 and call `fetch` with it.
-If the response is not ok, your function must throw an error using
-the response status message.
+If the response is not ok (i.e. response status is not 2xx), your function must throw an error using
+the HTTP status code message.
 
-The response body must then be read and parsed from json.
+The response body must then be read and parsed as json.
 
 The parsed object contains one of those 2 properties:
 - `"data"` the actual data to return


subjects/get-json/README.md
```diff
@@ -5,15 +5,15 @@
...
-- `path`, that will be the url called by your function
-- `params` *optional*, that will be the search parameters appended to your url
+- `path`, string that will be the url called by your function
+- `params` *optional*, object that will be the search parameters appended to your url
 
 `getJSON` must construct a valid url with the `path` and stringified `params`
 and call `fetch` with it.
-If the response is not ok, your function must throw an error using
-the response status message.
+If the response is not ok (i.e. response status is not 2xx), your function must throw an error using
```
kigiri, line 13: I was being evasive on purpose here, so student search a bit.
Also it was a hint on [response/ok](https://devdocs.io/dom/response/ok)

Which is cleaner to use than checking the range of `statusCode`.

I'm pretty sure everyone will just check status code directly if we write that.

subjects/get-json/README.md
```diff
@@ -5,15 +5,15 @@
 In this exercise, we will focus on building complex async flows with promises.
 
 Create a `getJSON` function that takes 2 parameters:
-- `path`, that will be the url called by your function
-- `params` *optional*, that will be the search parameters appended to your url
+- `path`, string that will be the url called by your function
+- `params` *optional*, object that will be the search parameters appended to your url
```
kigiri, line 9: Good precision

subjects/get-json/README.md
```diff
@@ -5,15 +5,15 @@
...
 
 `getJSON` must construct a valid url with the `path` and stringified `params`
 and call `fetch` with it.
-If the response is not ok, your function must throw an error using
-the response status message.
+If the response is not ok (i.e. response status is not 2xx), your function must throw an error using
+the HTTP status code message.
```
kigiri, line 14: I'm good with adding HTTP, but it's not the code if it's the message, it can't be both, a status as a code and a message not a code message.

Maybe specify the request status message ?

subjects/get-json/README.md
```diff
@@ -5,15 +5,15 @@
...
 
-The response body must then be read and parsed from json.
+The response body must then be read and parsed as json.
```
kigiri, line 16: thx for all the PR's @rtabulov

kigiri (changes requested):

PR #742, opened 2021-02-25
Pjs lite - nodejs quest

lfoussat:


subjects/tell-it-cypher/README.md
```diff
@@ -0,0 +1,18 @@
...
+### Instructions
+
+FINALLY, you got the names.
+But you could forget it... loose it... How could you write it without puting the surprise in danger?!
+
+Create a `tell-it-cypher.mjs` script that:
+- Takes a file as first argument
+- Takes one of these keywords as second argument:
+  - `encode`: convert your file from `utf8` to `base64`, then save the result in a `cypher.txt` file.
+  - `decode`: convert your file from `base64` to `utf8`, then save the result in a `clear.txt` file.
```
xpetit, line 12: Base64 is a binary to ASCII encoding, so it can take anything and will convert it to ASCII, which is still UTF-8.
So it's less confusing to not mention UTF-8 at all, only Base64.

PR #754, opened 2021-03-08
first pass remarks

Frenchris:
- Traits, 
Unclear,
"Define both structures fruits and meat", 
The structures are already defined in the subject

Maybe say "Define the impl food for fruit and for meat" 
Additionally

Fruits increases strength by 1 unit
Meat increases strength by 3 units

In the example given 
Player eats an apple and gets  strength increase by 4
Eats a steak and gets strength increased by 9

The solution give the food weight in kg multiplied by 4

There is a lack of information about the formula. This exercise is not achievable as it is. It needs clarifications.

Morevoer,

"Define the std::fmt::Display trait of the Player structure so using
// the template {} inside a println! macro will print in the first
// line the name of the player
// in the second line the strength, score and the money
// and in the third line the weapons"

Isn't the exercises missing this part of the tests?
I do not see any test about the display trait of the player structure
Nor do I see it implemented in the solution

- matrix and matrix_ops and matrix_mult 
The solutions on which the tests are based should be separated. 
There should a 
Matrix folder
Matrix_ops folder
And 
Matrix_mult folder
This will cause confusion in case of bugs for the person debugging during an ongoing piscine. Or an exam if those exercises are reused. 
Based on the separation of those tests I would need the working solutions as “students” to clarify the instructions making the “bridge” between exercises. (examplle: reuse matrix from previous exercise, import it? etc…) 

Those are not easy topics. The instructions have to be as clear as possible for those more than the simpler exercises. 

When correcting these please contact me so that we can have a discord share screen to show what i mean.  We will run them as student on the platform. 

- Lalgebra_vector
A vector in linear algebra is define as "anything that can be added and that can be multiplied by a scalar" 
May we have the source of this definition and add it to the subject?

Also, the instructions were missing the point of the exercise

Phrases starting with verbs such as...
Define this…
Define the function new() in the impl
Create this …
Make this work… 

Etc… 

What is the ~Rhs~ as opposite to Self?

I think this subject needs a full reread of the instructions. 

In addition,
If there is a use of the trait scalar coming from another exercise
IT HAS TO BE SPECIFIED in the subject.
The student can not be left to guess. 

- blood_types
Changed the name from enumerator to enum in subject. Check if I am correct.
Cool exercise, well written, maybe a bit long or it should be more towards the end of the quest.

- border_cross
Subject missing. No subject fetching. I did find the test but no subject with a similar name on it. 
TODO: 
This subject needs a run on the platform as student. The fetching of the subject needs to be checked.

- roman_numbers and roman_numbers_iter
Same remarks than for the matrix exercises 
Roman_number and roman_numbrs_iter should have separated solutions folders.
The solutions on which the tests are based should be separated. 
I do not know if I have to import from previous exercises or if I have to redefine.
In case of re import this needs to be specified!
When correcting these please contact me so that we can have a discord share screen to show what i mean. We will run them as student on the platform. 

- vector_operations
Object wrongly named in migration. vectors_operations. (Extra s)
Migration needed to fix or change of subject and test names.
Additionally,
Subject unclear.
Long sentence with no punctuation and an open parenthesis that does not end.
Please re read and fragment the part about what a 3 dimensional vector is. 

Notify me when Subject rewritten.
TODO: 
This subject needs a run on the platform as student

- delete_prefix
Object wrongly named in migration. vectors_operations. (Extra r and e reversed)
Migration needed to fix or change of subject and test names.

Subject seems clear. But maybe more examples are needed.
For instance what happens if a prefix I run 

   println!("{:?}", delete_prefix("bc", "abcdefghijklmnop"));

Is bc considered a prefix? Or do I get a none because of the a letter?
TODO: 
This subject needs a run on the platform as student

- commits_stats
Rust test with sol running and subject clear.
However when running example of usage locally after I added the fmt::*; import I get a backtrace error
`thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/main.rs:5:55
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace`

I need assistance with running the usage example as a student locally with the segregation of main.rs and lib.rs file.

### DIFF
--- a/subjects/blood_types/README.md
+++ b/subjects/blood_types/README.md
@@ -2,12 +2,13 @@
 
 ### Instructions
 
-In this exercise you will create a data model of blood types and an API to deal with blood types
+In this exercise you will create a data model of blood types and an API to deal with them.
 
 Start by copying the data representation of the blood types:
-- Create the enumerator `Antigen` that has 4 possibilities: A, B, O and AB And the enumerator `RhFactor` that has two possible values: Positive and Negative
 
-- After, copy the struct BloodType that contains two fields with the names antigen and rh_factor
+- Create the enum `Antigen` that has 4 possibilities: A, B, O and AB And the enum `RhFactor` that has two possible values: Positive and Negative
+
+- After, copy the struct `BloodType` that contains two fields with the names antigen and rh_factor
 
 - To provide a simple way to create blood types implement the trait FromStr for BloodType (which will allow us to use the `parse` method and the associated function from_str, so we can do:
 
@@ -85,6 +86,8 @@ impl BloodType {
 Here is a program to test your function.
 
 ```rust
+use blood_types::*;
+
 fn main() {
 	let blood_type: BloodType = "O+".parse().unwrap();
 	println!("recipients of O+ {:?}", blood_type.recipients());
--- a/subjects/border_cross/README.md
+++ b/subjects/border_cross/README.md
@@ -1,0 +1,79 @@
+## vehicles
+
+### Instructions
+
+- Create the trait Vehicle with the model and year
+
+- In a border cross you want to keep a list of all the vehicles that are waiting to enter the country. You want to keep a waiting list of the vehicle but the vehicles can be of two types: Car or Truck
+
+- Create a function that receives a vector of structures that implements the Vehicle trait and returns all the models.
+
+With the following structure:
+
+### Expected Functions and Structures
+
+```rust
+#[allow(dead_code)]
+pub struct Car<'a> {
+	pub plate_nbr: &'a str,
+	pub model: &'a str,
+	pub horse_power: u32,
+	pub year: u32,
+}
+
+#[allow(dead_code)]
+pub struct Truck<'a> {
+	pub plate_nbr: &'a str,
+	pub model: &'a str,
+	pub horse_power: u32,
+	pub year: u32,
+	pub load_tons: u32,
+}
+
+pub trait Vehicle {
+	fn model(&self) -> &str;
+	fn year(&self) -> u32;
+}
+
+impl Vehicle for Truck<'_> {
+}
+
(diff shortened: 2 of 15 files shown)


augusto-mantilla:
I think there was a problem at some point. There was a problem at some point those are not the final version of the readmes. I'm really confused. Maybe I messed up a rebase. I'm going to fix them soon.

augusto-mantilla:
About importing `fmt::*`. NEVER do that! Importing everything is only reserved for tests. Check again in the book explains why.

augusto-mantilla:
About creating a different folder for the exercises defining traits. It's not alllowed to define traits for types that are not declare un the same crate. It's called the 'orphan rule'. You can read more about it in the book. That's the reason it has to be in the same folder of other exercises.

augusto-mantilla:
About `Rhs`. If you look for the traits in `std::ops`. Like for example `std::ops::Add` there is an explanation of what it is. And how to use it.

augusto-mantilla:
@Frenchris  I found the subject for border_cross! :smile:

augusto-mantilla:
@Frenchris I fixed you're remarks about the trait exercise. Now it should be better.

augusto-mantilla:
@Frenchris About the definition of vector I took it from the wikipedia page https://en.wikipedia.org/wiki/Vector_space, but it has some much math that I'm afraid that people are going to get scare and I only care about that detailed. "Anything that can be added and that can be multiplied by a scalar". There is a nice translation from that sentence to a simple trait bound.

augusto-mantilla:
@Frenchris about the exercise delete_prefix I can add that extra example that you suggest, but for me it's pretty clear that it will return `None` since "bc" **is not** a prefix for "abcdefghijklmnop".

augusto-mantilla (approved):

PR #786, opened 2021-06-02
test and resources check quest-02 and quest-03

LEEDASILVA:
1. adding of resources :
- exercises using `PrintRune` will have in notions : 
  - `[01-edu/z01](https://github.com/01-edu/z01)`

- exercise `printnbr` will have the additional resource:
  - `[numeric types](https://golang.org/ref/spec#Numeric_types)`, this resource is added because the subject restricts the student from using other `int64`

- exercises that use pointers will have the following resource:
  - `[Pointers](https://golang.org/ref/spec#Pointer_types)` was thinking of this one https://tour.golang.org/moretypes/1. Let me know what you think @xpetit 

- exercises that use `atoi` will have the resource : `[strconv/Atoi](https://golang.org/pkg/strconv/#Atoi)`

-  other exercises from the quest 02 and 03 did not need resources, videos already explain the basics.
  in case we need the resources i was thinking of using the golang.org. for example: `[if statement](https://golang.org/ref/spec#If_statements)`.
  let me know what you think @xpetit

2. tests:
- all go tests where tested using the platform and locally

- found some code that is not being used in `tests/pointone_test/main.go` did change it, but after doing a pr i wanted to now if it is normal.
- added a line on the subject `printcombn`, where students must be aware of the code performance: `> Be mindful of your program efficiency to avoid timeouts.`


xpetit:
>   * `[Pointers](https://golang.org/ref/spec#Pointer_types)` was thinking of this one https://tour.golang.org/moretypes/1. Let me know what you think @xpetit
Yeah Go tour is good !

xpetit (approved):

augusto-mantilla (approved):

PR #790, opened 2021-06-11
1st Stats Projects

OGordoo:
This PR contains the READMEs and the audits of the new projects requested by Nicolas.
It consists of 4 projects in which the math knowledge of the students will be put to the test. These projects focus on statistical calculations, like the average, standard deviation, linear regression line, etc.
The first project focus on the calculation of the average, median, standard deviation and variance of a set of data. Similarly the third exercise consists on the calculation of the linear regression line and the Pearson's correlation coefficient. The second and fourth projects consist on the usage of the previous calculus to guess a next number of a sequence of numbers.

### DIFF
--- a/subjects/guess-it-1/README.md
+++ b/subjects/guess-it-1/README.md
@@ -1,0 +1,76 @@
+## guess-it-1
+
+### Objectives
+
+Now that you have the math skills up and running , it is time for you to guess numbers. Find a way to implement the `math-skills` exercise into this one.
+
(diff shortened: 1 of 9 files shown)


xpetit:
Awesome!
Where is the source code for https://assets.01-edu.org/stats-projects ?

Also with `GOOS` env variable you can compile macOS & windows binaries

Frenchris (approved):

PR #793, opened 2021-06-21
printmemory Change exercise

OGordoo:


subjects/printmemory/README.md
```diff
@@ -4,9 +4,9 @@
 
 Write a function that takes `(arr [10]int)`, and displays the memory as in the example.
```
xpetit, line 5: Should be `byte` instead of `int`, and a slice instead of an array, test needs to be updated.

This way in the example you can use byte literals instead of int values:
```go
PrintMemory([10]int{104, 101, 108, 108, 111, 16, 21, 42})
```
=>
```go
PrintMemory([...]byte{'h', 'e', 'l', 'l', 'o', 16, 21, '*'})
```

xpetit (changes requested):

xpetit (approved):

PR #804, opened 2021-07-13
go-reloaded Refactor

OGordoo:


Frenchris:
reviewed, but not merging it yet. Awaiting some thime that the new alem promotion pass over the old version.

PR #807, opened 2021-07-15
add missing allowed packages

alseiitov:
added missing allowed packages for forum optionals, real-time-forum and real-time-forum typing-in-progress;
moved allowed packages section to the end of readme file in forum moderation, like in other readme files


OGordoo:
Thank you for your changes 👍

PR #814, opened 2021-07-30
Add new subject for the map exercice in the piscine

Clemalfroy:
Tested  and written with @Frenchris

### DIFF
--- a/subjects/maxwordcountn/README.md
+++ b/subjects/maxwordcountn/README.md
@@ -2,7 +2,7 @@
 
 ### Instructions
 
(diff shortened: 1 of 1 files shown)


subjects/maxwordcountn/README.md
```diff
@@ -0,0 +1,41 @@
+## maxwordcountn
+
+### Instructions
+
+Write a function `MaxWordCountN` that will return a `map` of the `n` words that occurs the most in a string `text`. This map will have the word as key and the number of occurences of this word as value.
```
xpetit, line 5: Maybe words should be explicitly defined, that they are separated by spaces

xpetit (changes requested):
Good subject for maps

subjects/maxwordcountn/README.md
Clemalfroy: Good idea ! I have added that thx

xpetit (approved):

PR #821, opened 2021-08-26
fixing ascii art

MSilva95:


Zhadyra21:
Hello! "The program always prints a new line at the end" However, looking at the examples, it prints not a new line, but several lines. If we run **fmt.Print("Hello")** without and with an extra line, it prints
ascii-art % go run . | cat -e      **(without \n)**
Hello%                                                                                                       
ascii-art % go run . | cat -e      **(with \n)**
Hello$
ascii-art % 
So I would expect the same in ascii-art. Not
ascii-art % go run . | cat -e
**Hello$
$**
Otherwise, "Hello\nWorld" should print in art:
Hello (in art)
$ (for **\n**)
World (in art)
$ (for the last line that is always printed)

LEEDASILVA:
@Zhadyra21 so if you take a look into the `files changed` you can see that in the `usage` we give an example.
```sh
student$ go run . "Hello\n" | cat -e
 _    _          _   _          $
| |  | |        | | | |         $
| |__| |   ___  | | | |   ___   $
|  __  |  / _ \ | | | |  / _ \  $
| |  | | |  __/ | | | | | (_) | $
|_|  |_|  \___| |_| |_|  \___/  $
                                $
                                $
$
$
``` so there is two new lines

LEEDASILVA:
@Zhadyra21 so you can take a look in the  `changed files` and in the `usage` & `audit`s we changed the examples.

PR #845, opened 2021-10-14
Update subjects/islower/README.md

Pav01Founders:
Updated instructions to fix grammar.

### DIFF
--- a/subjects/islower/README.md
+++ b/subjects/islower/README.md
@@ -2,7 +2,7 @@
 
 ### Instructions
 
-Write a function that returns 'true' if the 'string' passed as the parameter contains only lowercase characters; returns 'false' otherwise.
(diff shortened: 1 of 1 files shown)


subjects/islower/README.md
```diff
@@ -2,7 +2,7 @@
 
 ### Instructions
 
-Write a function that returns `true` if the `string` passed in parameter only contains lowercase characters, and that returns `false` otherwise.
+Write a function that returns 'true' if the 'string' passed as the parameter contains only lowercase characters; returns 'false' otherwise.
```
OGordoo, line 5: instead of the `;` we prefer the `and that`

HanaAbdi: "and that" doesn't make sense in the english language.   A semicolon is used to join two related independent clauses in place of a comma and a coordinating conjunction.

PR #848, opened 2021-10-14
Update subjects/isprintable/README.md

Pav01Founders:


subjects/isprintable/README.md
```diff
@@ -2,7 +2,7 @@
 
 ### Instructions
 
-Write a function that returns `true` if the `string` passed as a parameter only contains printable characters, and returns `false` otherwise.
+Write a function that returns `true` if the `string` passed as a parameter contains only printable characters; returns `false` otherwise.
```
OGordoo, line 5: instead of the `;` we prefer the `and that`

PR #878, opened 2021-10-31
Fix console error message

nik-don:
Removed additional unwanted text.


MSilva95:
This was solved, thank you for the catch!

PR #885, opened 2021-11-03
Rephrase the question (ascii-art audit)

nik-don:
Currently the question is confusing to read - It is like the second part is negating the first question.


LEEDASILVA:
Thank you for the PR, it will be merged

PR #887, opened 2021-11-10
Add package name (piscine) to test function

nik-don:
piscine imported but not being used in test function


LEEDASILVA:
thank you for the correction

Issue #905, opened 2021-12-13
forum-security misleading info


GermanKarK:
Students wrote about some misleading info with the forum project which are shown in the screenshots
![forum-security_1](https://user-images.githubusercontent.com/95408096/145812156-7c233af4-a381-4894-ac9f-a96ed558ba9e.png)
![forum-security_2](https://user-images.githubusercontent.com/95408096/145812167-a7ca765c-9e0b-4438-a271-86ba3565ad36.png)

LEEDASILVA:
@GermanKarK thank you for the feedback.
session and cookies are to be implemented in the mandatory project. But not tested in a secure way! (in the audit) So this repetition of session and cookies is so that students can actually see that there are possibilities for attacks and correct them in this project.

The password encryption is repeated  as well because it is not tested in the audit. This project would insure the encryption.

`- It is recommended that the code should present a **test file**.`-> the phrase is not well formed i will correct it. But in programing it is very very important that test are made. Testing is any activity aimed at evaluating an attribute or capability of a program or system and determining that it meets its required results. So students must do a test files

is it clear?

uristdwarf:
Hey, I'm the author that image.
I agree that cookies and sessions can be implemented in a bad way, however the main project's description already strongly advises you to implement these, so I guess it would be better if it doesn't and later in security this is introduced?

As for testing, I'm assuming you mean unit testing? In this case it could be something like "Is the code covered by [unit tests](https://go.dev/doc/tutorial/add-a-test)?" or something like that

The cipher suites are indeed already provided by most HTTPS certificates and CA's, which is why I find them confusing. Do we need to choose cipher suites the client can use, or are the default fine?

LEEDASILVA:
I do not know if the best option is to remove the session & cookies from the main subject. But maybe we could add it was a bonus, so students can decide if they do it or not. And when arriving to security the students must do it in the right way. what do you think?

Yes, unit tests (for the server side), should have been more clear on this one. **But will add the specifications**

For the cipher suites, the default are fine but the idea is for students to search and know about them!
for example you can use the `tls.Config` and enable the `PreferServerCipherSuites` -> this causes server to use Go's default cipher suites preferences, which are tuned to avoid attacks.

uristdwarf:
I think that securing cookies/sessions/passwords could be indeed bonus tasks. although I think sessions will still need to be implemented (even if insecure) in the main project

[...]

PR #907, opened 2021-12-13
forum/security : issue correction

LEEDASILVA:
issue https://github.com/01-edu/public/issues/905


uristdwarf:
Also a lot of the other subjects will have to make it clear that "test file" means "unit testing", but that can be a separate issue/pull request.

LEEDASILVA:
@uristdwarf all done.

Next time do not **hesitate to make a PR**. After all making PRs is part of being a programmer, and a good way to learn is by doing.

PR #936, opened 2022-01-21
fusion: More logical ordering of instructions

uristdwarf:
Had some confusion about what it meant by type mismatch when it contained 2 values whose type are the same (both are objects), but when reading the next instruction first (about recursion) it's a lot more clearer.


MSilva95:
Hello @uristdwarf  this is fixed in this pr: https://github.com/01-edu/public/pull/938 
i was already making the changes when you created the pr but thank you for your efforts. I will take a look at your other prs as well

PR #967, opened 2022-02-17
pick-and-click test fix

MSilva95:


maximihajlov:
Fix doesn't work. The clipboard override should be before `page.mouse.click(...move)`. Now when clicking, `writeText` and `readText` are not redefined yet, so nothing is written to `clipboardText` and `readText` returns real clipboard contents.

maximihajlov:
Also, I don't know why, but your fix doesn't work for me. But this one works:
```js
let clipboardText = null;
window.navigator.clipboard.readText = () => new Promise(resolve => resolve(clipboardText))
window.navigator.clipboard.writeText = (text) => new Promise(() => clipboardText = text)
```

PR #971, opened 2022-02-18
public issues

MSilva95:


dom/pick-and-click_test.js
```diff
@@ -77,16 +77,17 @@ tests.push(async ({ page, eq, bodyBgRgb, random }) => {
...
       let clipboardText = null
-      window["navigator"]["clipboard"] = {
-        writeText: (text) => new Promise((resolve) => (clipboardText = text)),
-        readText: () => new Promise((resolve) => resolve(clipboardText)),
-      }
-      return navigator.clipboard.readText()
+      window.navigator.clipboard.readText = () =>
+        new Promise((resolve) => resolve(clipboardText))
+      window.navigator.clipboard.writeText = (text) =>
+        new Promise(() => (clipboardText = text))
```
kigiri, line 85: I think we have an issue here as we do not ever terminate this promise, this code will fail if the user try to `await` the `writeText` to ensure the write worked, which would be the proper solution.

A cleaner and, arguably simpler fix would be:
```js
window.navigator.clipboard.readText = async () => clibboardText
window.navigator.clipboard.writeText = async (text) => {
  clipboardText = text
}
```

PR #981, opened 2022-02-20
update rust readme quest 01 to quest 03

davidrobert99:


subjects/bigger/README.md
```diff
@@ -6,11 +6,7 @@ Create the function `bigger` that gets the biggest positive number in the `HashM
 
 ### Notions
 
-[hash maps](https://doc.rust-lang.org/book/ch08-03-hash-maps.html)
-
-### Dependencies
-
-rand = "0.7"
+- [hash maps](https://doc.rust-lang.org/book/ch08-03-hash-maps.html)
```
LEEDASILVA, line 10: why remove the dependencies?
we use the version of the dependencies because of some problems we encountered.
Did it pass without using this version?

subjects/borrow_me_the_reference/README.md
```diff
@@ -28,8 +28,8 @@ pub fn is_correct(v: &mut Vec<&str>) -> usize {
 
 ### Notions
 
-- https://doc.rust-lang.org/book/ch04-00-understanding-ownership.html
-- https://docs.rs/meval/0.2.0/meval/
+- [ownership](https://doc.rust-lang.org/book/ch04-00-understanding-ownership.html)
+- [meval](https://docs.rs/meval/0.2.0/meval/)
```
LEEDASILVA, line 32: nice consistency ! haha

LEEDASILVA (commented):
The only issue I'm having is the version removal the rest looks good

subjects/bigger/README.md
davidrobert99: Yeah, I didn't have to use rand in that exercise, so I thought that it was a mistake. The exercise is to get the biggest positive number in a HashMap, was I supposed to use rand for that?

PR #1003, opened 2022-03-07
Update README.md for atm management audit

EvilgeniuS1982:


subjects/atm-management-system/audit/README.md
```diff
@@ -6,73 +6,73 @@
...
 
-##### Resorting to the user Alice select the option `"Make transaction"`, choose the account with the id `"3212"`. Then try to withdraw money.
+##### Referring to the user Alice select the option `"Make transaction"`, choose the account with the id `"3212"`. Then try to withdraw money.
 
 ###### Are you able to withdraw money?
 
 ###### And if so, was the withdraw updated in the file `"records.txt"`?
 
-###### Is it not possible to withdraw an amount superior to your available balance?
+###### Is it possible to withdraw an amount superior to your available balance?
```
MSilva95, line 62: If you remove the "not" it will change the sentence.
in order to pass the questions the answer for it should be always "yes", otherwise it will fail the audit.
and if you say  yes to this question you are saying that it is possible to withdraw an amount superior to your available balance which should not be possible!
We can do it like this:

###### Can you confirm that it is not possible to withdraw more than your available balance?

Now if you say yes it is correct and it will pass the audit

subjects/atm-management-system/audit/README.md
```diff
@@ -6,73 +6,73 @@
...
 
-###### Are all the user's names unique? (ex: no repetition on the name Alice)
+###### Are all the user names unique? (ex: no repetition on the name Alice)
 
 ##### Try and login as `"Alice"`.
 
 ###### Was Alice able to enter the main menu?
 
-##### Try to create two accounts using the user Alice, then select the option `"Update information of account"` and select an account number that does not exist for Alice.
+##### Try to create two accounts with the user Alice, then select the option `"Update account information"` and select an account number that does not exist for Alice.
```
MSilva95, line 19: This is fine!
just want to let you know that i already changed it in the project that is given to the students as well.

subjects/atm-management-system/audit/README.md
```diff
@@ -6,73 +6,73 @@
...
 
-###### Did the application displayed some kind of error message stating that this account does not exist?
+###### Did the application display some kind of error message stating that this account does not exist?
 
-##### Resorting to the user Alice, try and select the option `"Update information of account"` and select one of the accounts you created for Alice.
+##### Referring to the user Alice, select the option `"Update account information"` and select one of the accounts you created for Alice.
```
MSilva95, line 23: The "referring" is not the best option i think "Using" is better
Because resorting and using is basically the same thing but  referring is different

subjects/atm-management-system/audit/README.md
```diff
@@ -6,73 +6,73 @@
...
 
 ###### And if so did it update the file `"records.txt"`?
 
-##### Resorting to the user Alice try to select the option `"Remove existing account"` and remove the accounts `"834213"`, `"320421"` and `"3214"`.
+##### Referring to the user Alice try to select the option `"Remove existing account"` and remove the accounts `"834213"`, `"320421"` and `"3214"`.
 
 ###### Can you confirm that those account were deleted, both in the application and file `"records.txt"`?
 
-##### Resorting to the user Alice select the option `"Remove existing account"` and try to remove and nonexisting account.
+##### Referring to the user Alice select the option `"Remove existing account"` and try to remove a account that doesn't exist.
```
MSilva95, line 73: we try to avoid using the apostrophe to make the English simple and easy for every one, because not all the students have good English. So, you need to avoid using (don't, doesn't, ...) and instead use (do not, does not, ...)

##### Using the user Alice select the option `"Remove existing account"` and try to remove an account that does not exist.

subjects/atm-management-system/audit/README.md
```diff
@@ -82,13 +82,13 @@
...
 
-###### +Was the user whom received the account notified instantly?
+###### +Was the user who received the account notified instantly?
 
 ###### +Did the student update the terminal interface?
 
 ###### +Is the password saved in the file `"users.txt"` encrypted?
 
-###### +Did the student create a relational database? • 6
+###### +Did the student create a relational database?
```
MSilva95, line 91: The "• 6" must not be removed.
It is part of the structure of the audit, it is a way to give weight to the bonus questions so that we can give more or less XP points

MSilva95 (changes requested):
I added some comments for you to change some stuff, because some things may not be changed.
You have to be careful not to change the sentence of the questions in the audits

subjects/atm-management-system/audit/README.md
Frenchris: We should maybe change the sentence to: `Does the system forbid to withdraw an amount superior to your available balance`?

subjects/atm-management-system/audit/README.md
Frenchris: should not it be `remove **an** account that does not exist`

MSilva95:
solved in #2195

PR #1036, opened 2022-03-25
issues on public

MSilva95:


sh/tests/to-git-or-not-to-git_test.sh
```diff
@@ -4,8 +4,16 @@
...
 
-submitted=$(bash student/to-git-or-not-to-git.sh)
-expected=$(bash solutions/to-git-or-not-to-git.sh)
-
-diff <(echo "$submitted") <(echo "$expected")
+if [ -f ${FILENAME} ]; then
+    if [ -s ${FILENAME} ]; then
+        submitted=$(bash student/to-git-or-not-to-git.sh)
+        expected=$(bash solutions/to-git-or-not-to-git.sh)
+        diff <(echo "$submitted") <(echo "$expected")
```
LEEDASILVA, line 13: it will still pass with a wrong solution. Maybe the problem will be the solution of this test

PR #1038, opened 2022-03-28
instructions for the admins to test the raids

MSilva95:


subjects/sudoku/audit/README.md
```diff
@@ -1,5 +1,7 @@
 #### Sudoku
 
+***NOTE: If you are an admin and you want to test this project, follow the instructions [HERE](https://github.com/01-edu/go-tests)***
```
kigiri, line 3: Did you check how this looked in the platform ?

subjects/quad/README.md
```diff
@@ -1,4 +1,4 @@
-# quadrangle
+## quadrangle
```
kigiri, line 1: why did you change this ? Is it for display reason on the platform ?

Structurally it made more sense for it to be H1, maybe we should lower the following titles to H3 instead of H2

MSilva95: I changed it to be the same as the other projects, for consistency, but if the structure is not relevant in this case we can let it be an H1 just to distinguish the main tittle better

MSilva95: quad-readme

This is what it will look like

kigiri (approved):

PR #1044, opened 2022-04-05
Update README.md

October45:
Changed the content of the file created to be noted as a codeblock instead of with quotation marks to make the contents of the file easier to understand. (Misinterpreted as including the quotation marks)


MSilva95 (approved):
This is a good idea! Thank you for the tip!

PR #1046, opened 2022-04-05
who-are-you extra explanation

sagarishere:
extra explanation proposed for expected solution:
as people who are familiar with jq are in habit of using raw output -r
however, this one expects a string output


MSilva95:
Hello @sagarishere! 
I have allready added some changes to the readme here:
<https://github.com/01-edu/public/pull/1050/files>
Thanks for the pr i will close it just because of the structure used but it was very helpful!

PR #1049, opened 2022-04-08
generics list exercise

davidrobert99:


davhojt (approved):
I'm unsure we need to be providing example main functions at this stage of the piscine, but
@davidrobert99 explained that the rest of the exercises also contain mains. We will review this later.

PR #1073, opened 2022-05-05
updating the chaikins raid

MSilva95:


davhojt (approved):
There's one or two grammar improvements to be done, but @davhojt will start a language review of the whole Rust Piscine next week.

PR #1084, opened 2022-05-17
Update roman numbers iter and rpn c 555

davidrobert99:


MSilva95 (changes requested):
just remove the extra lines and  fix the quotes at the end of rpn and its a go.

MSilva95 (approved):

PR #1091, opened 2022-05-23
1 review language piscine rustquest 02 c 461

davhojt:
Language review of the second rust piscine.

### DIFF
--- a/subjects/copy/README.md
+++ b/subjects/copy/README.md
@@ -15,6 +15,7 @@ Create the following **functions**. The objective is to know how ownership works
   - with the `original` value.
   - and the `natural logarithm` of each `absolute` value.
 
+### Expected functions
 ```rust
 pub fn nbr_function(c: i32) -> (i32, f64, f64) {
 }
--- a/subjects/copy/README.md
+++ b/subjects/copy/README.md
@@ -2,27 +2,18 @@
 
 ### Instructions
 
-Your objective is to fix the **functions** work so that the program works.
+Create the following **functions**. The objective is to know how ownership works with different types.
 
 - `nbr_function` returns a tuple:
-  - with the `original` value
-  - the `exponential function` of the value
-  - and the `natural logarithm` of this `absolute` value
+  - with the `original` value.
+  - the `exponential function` of the value.
+  - and the `natural logarithm` of the `absolute` value.
 - `str_function` returns a tuple:
-  - with the `original` value
-  - and the `exponential function` each value as a string (see the example)
+  - with the `original` value.
+  - and the `exponential function` of each value as a string (see the example).
 - `vec_function` returns a tuple:
-  - with the `original` value
-  - and the `natural logarithm` of each `absolute` value
-
-The objective is to know how ownership works with different types.
-
-### Notions
-
-- [scope](https://doc.rust-lang.org/rust-by-example/scope/move.html)
-- [Primitive Type f64](https://doc.rust-lang.org/std/primitive.f64.html)
-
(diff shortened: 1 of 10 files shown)


subjects/copy/README.md
```diff
@@ -2,27 +2,18 @@
...
-  - and the `natural logarithm` of each `absolute` value
-
-The objective is to know how ownership works with different types.
-
-### Notions
-
-- [scope](https://doc.rust-lang.org/rust-by-example/scope/move.html)
-- [Primitive Type f64](https://doc.rust-lang.org/std/primitive.f64.html)
-
-### Expected Functions
```
MSilva95, line 25: Why did you remove the "###Expected functions"?

subjects/doubtful/README.md
```diff
@@ -2,23 +2,15 @@
...
-
-You have to complete the signature the functions to make it fit the usage and you also have to modify the usage to be able to do what is expected.
-
-- Restrictions:
-
-  - `doubtful` cannot return any value.
-
-  - And in the example of the usage you can only modify the argument of the function
-
-### Expected
```
MSilva95, line 15: Same here, why not having the "expected files"?

subjects/string_literals/README.md
```diff
@@ -4,21 +4,11 @@
...
-
-- [stack and heap](https://doc.rust-lang.org/1.22.0/book/first-edition/the-stack-and-the-heap.html)
-- [Literals](https://doc.rust-lang.org/rust-by-example/primitives/literals.html)
-
-### Expected Functions
+- `is_empty`: that returns `true` if the string is empty.
+- `is_ascii`: that returns `true` if all characters are within the ASCII range.
+- `contains`: that returns `true` if the string contains the given pattern.
+- `split_at`: that divides a string in two returning a tuple.
+- `find`: that returns the index if the first character of a given string that matches the pattern.
```
MSilva95, line 11: `find`: that returns the index of the first character of a given string that matches the pattern.

MSilva95 (changes requested):
There are some things to change, but my biggest confusion is why you removed the Expected Functions, as this is the structure used in the other piscines as well

subjects/string_literals/README.md
davhojt: Cheers, fixed.

subjects/tic_tac_toe/README.md
```diff
@@ -2,33 +2,18 @@
...
-
-### Expected Functions
+Create a function named `tic_tac_toe`, which receives a tic-tac-toe table. It should return the appropriate string: `"player O won"`, `"player X won"` or `"Tie"`.
 
 ```rust
 pub fn tic_tac_toe(table: Vec<Vec<&str>>) -> String {
 }
+```
+
+Also create the following functions, which each accept a player and a table. These functions should return `true` if the player has completed one of the diagonals, rows or columns:
```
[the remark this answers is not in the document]
davhojt: It's back in it's normal order, so that there is only one inline block.

MSilva95 (approved):

davhojt:
We can discuss them later as a group.

I'd consider even removing function signatures entirely in some cases, like in the JS piscine. To not remove learning opportunities. I'd predict that many students just use ours, instead of considering their own parameter naming for example.

I'm also not sure about structure. On the one hand, consistently is beneficial for ease of browsing. But students also must become skilled at reading multiple types of documentation - and learn to read between the lines. But we can talk about this as a group too.

Cheers for the feedback, I've implemented the changes 👍🏾.

PR #1092, opened 2022-05-24
Update README.md for doubtful subject

maximihajlov:
Add doubtful package import statement

### DIFF
--- a/subjects/doubtful/README.md
+++ b/subjects/doubtful/README.md
@@ -17,6 +17,8 @@ You'll need to complete the function signature, so that it works properly with t
 Here is a program to test your function
 
 ```rust
+use doubtful::*;
+
 fn main() {
 	let mut s = String::from("Hello");
 


subjects/doubtful/README.md
```diff
@@ -17,6 +17,8 @@ You'll need to complete the function signature, so that it works properly with t
 Here is a program to test your function
 
 ```rust
+use doubtful::*;
```
davhojt, line 20: The usage is intentionally incomplete here. The subject includes the line:
> You'll also need to complete the usage if you plan to use it.

Perhaps as it is the first quest, we can include a hint that the something needs to be done. For example, the function call has missing arguments, so we added a comment:
```rust
doubtful(/*add your code here*/);
```

Let's do something similar with this. Something like:
```rust
use /* add your code here */
```

davhojt (changes requested):

davhojt:
Thanks @maximihajlov for your PR.

I've made a small proposal there to ensure that we provide learning opportunities to students. Take a look, and we welcome your feedback.

PR #1097, opened 2022-05-25
Update README.md for changes subject

maximihajlov:
Link to Rust doc fixed, old link https://doc.rust-lang.org/std/keyword.Self.html is 404.


MSilva95 (approved):
Thank you for the fix!

PR #1102, opened 2022-05-28
Fix expected function in description

AaEnnDeeErrEeEss:
Change expected function signature to match required result


AaEnnDeeErrEeEss:
Pull request with the fix already exists.

PR #1105, opened 2022-05-31
7 review language piscine rustquest 06 c 367

davhojt:
Language review of the second quest of the second week.

### DIFF
--- a/subjects/scores/README.md
+++ b/subjects/scores/README.md
@@ -2,11 +2,11 @@
 
 ### Instructions
 
-Lets play a little!
-Create a function `score` that given a string, computes the score for that given string.
+Lets play a little.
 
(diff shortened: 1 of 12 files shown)


subjects/scores/README.md
```diff
@@ -20,17 +20,8 @@ You will need these:
 | J, X                         |   8   |
 | Q, Z                         |  10   |
 
-### Notions
-
-- [patterns](https://doc.rust-lang.org/book/ch18-00-patterns.html)
-
 ### Expected functions
```
MSilva95, line 23: If we decide to remove the function here we need to remove the "### Expected Functions" as well

davhojt: Replaced my comment, as I had accidentally started a review previously.

The following is below:
> You'll need to work out the function signature for yourself.

PR #1109, opened 2022-06-01
docs(lunch_queue) correct grammar

davhojt:
Language review of a single exercise lunch_queue

It has been done alone, as it was not done as part of a previous language review. It is now only in exam >3

### DIFF
--- a/subjects/lunch_queue/README.md
+++ b/subjects/lunch_queue/README.md
@@ -2,19 +2,17 @@
 
 ### Instructions
 
-You will need to create an *API* which will organize a queue of people, so that a program can organize a queue of people.
+You will need to create an *API*, so that a program can organize a queue of people.
 
-The program requires the following functionsAdd the following associated functions to the `Queue` structure:
+The program requires the following functions. Add them as associated functions to the `Queue` structure:
 
 - `new`: which will initialize the `Queue`.
-- `add`: which receives a person's information, to add them to the `Queue`.
+- `add`: which adds a person to the queue.
 - `invert_queue`: which reverses the queue.
-- `rm`: which will remove the person who finished ordering their food. The removal should respect the FIFO method (first in first out. The function should return the removed person.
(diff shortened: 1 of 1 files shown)


subjects/lunch_queue/README.md
```diff
@@ -2,27 +2,19 @@
...
-- `search`, which returns a tuple with the information of a given person `id`
+- `new`: which will initialize the `Queue`.
+- `add`: which adds a person to the queue.
+- `invert_queue`: which reverses the queue.
+- `rm`: which removes the person who finished ordering their food. The removal should respect the FIFO method (first in first out). It should return the person's details.
+- `search`: which returns the details for a given person's `name`.
 
-You must also create a type called `Link`. This will be the connection of the structures `Queue` and `Person`.
-Do not forget that this will be a recursion type and it must point to `None` if there is no persons.
+You must also create a type named `Link`. This will be the connection of the structures `Queue` and `Person`. This will be a recursion type and it must point to `None` when the person is the last person in the queue.
```
MSilva95, line 15: Is it in the last person or when there are no more people?

MSilva95 (changes requested):
Just need to confirm if the line 15 is correct or not

subjects/lunch_queue/README.md
davhojt: Ah, I think you're right to point this out, as I guess we are talking about the `Queue` here, not `Person`.
The `Link` in `Queue` should point to a `Person` or `None` if there are no people.
The `Link` in `Persons` which we don't actually describe here should point to `None` when it is at the end of the *linked list* of people.

davhojt: I changed it here to make it a little more generic for `Queue` *and* `Person`: https://github.com/01-edu/public/pull/1109/commits/84226861cfa1f2116644ccf5479f68da1b6e560c

MSilva95: Cool

MSilva95 (approved):

PR #1115, opened 2022-06-03
fix: printhex README.md

Dias1c:
## The description did not fit the tester.
Here is an example of the error
Photo of the problem

Task description changed


MSilva95 (commented):
Your exercise is wrong!
If you look closely you can see that in the readme the example uses only one argument ```" 123 132 1" ``` and the test uses three arguments ``` "123" "132" "1"``` so they are two completely different things.

PR #1168, opened 2022-06-07
 buzzinga subject 

zainabdnaya:


Hamzaelkhatri (commented):
All is good.

subjects/buzzinga/README.md
```diff
@@ -0,0 +1,53 @@
...
+)
+
+func main() {
+	fmt.Println(BuzZinga(15))
+}
+```
+And its output :
+
+```go
+$ go run . | cat -e
```
MSilva95, line 36: What output is this?
If you pass 15 isn't the output just "BuzZinga" as the number is divisible by 3 and 5?

subjects/buzzinga/README.md
```diff
@@ -0,0 +1,53 @@
+# buzzinga
+
+### Instructions
+Write a function named `BuzZinga()` that takes a number as an argument and prints the following:
+
```
MSilva95, line 5: You need to say what the number 0 does...

MSilva95: and how negative works

MSilva95 (changes requested):

subjects/buzzinga/README.md
zainabdnaya: done

subjects/buzzinga/README.md
zainabdnaya: I didn't quite understand you.

MSilva95: my question is what is supposed to be the output for the number 15?
In my perspective, as it is divisible by 3 and 5 the output should be just  "BuzZinga"

subjects/buzzinga/README.md
```diff
@@ -0,0 +1,70 @@
+# buzzinga
+
+### Instructions
+
+Write a function named `BuzZinga()` that takes a number as an argument, then  loop over it and prints the following:
```
MSilva95, line 5: Write a function called `BuzZinga()` that takes a number as an argument, then loops over it and does the following for each iteration:

subjects/buzzinga/README.md
MSilva95: Its a bit confusing try changing the head wit this>

Write a function called `BuzZinga()` that takes a number as an argument, then loops over it and does the following for each iteration:

zainabdnaya: Done

subjects/buzzinga/README.md
```diff
@@ -0,0 +1,70 @@
+# buzzinga
+
+### Instructions
+
+Write a function called BuzZinga() that takes a number as an argument, then loops over it and does the following for each iteration:
```
MSilva95, line 5: Where is the Highlights in the func?

MSilva95 (changes requested):

MSilva95 (approved):

PR #1182, opened 2022-06-09
update tron

davidrobert99:


davhojt (commented):
I made osme small changes as @davidrobert99 is away. I'll merge and test in the platform.

davhojt (approved):
I made some small changes as @davidrobert99 is away. I'll merge and test in the platform.

PR #1184, opened 2022-06-09
add(subject):getarea

Hamzaelkhatri:


zainabdnaya (commented):
Is the return should be only integer or float and double.

Hamzaelkhatri:
@zainabdnaya  should print only integer, by casting the result from `float` or `double` to `int`

zainabdnaya (approved):

subjects/getarea/README.md
```diff
@@ -0,0 +1,25 @@
...
+    - if the number of arguments is not `1` print (`'\n'`)
+    - You may assume that the value of `π` is 3.14
+
+###  Usage
+
+```console
+$ go run . | cat -e
+$
+$ go run . 10 | cat -e
+315$
```
MSilva95, line 18: isn't this 314??

Also you need to explain that we want the output to be an integer and the proper way to round it

MSilva95 (approved):

Issue #1157, opened 2022-06-07
is Same String


mohamedLazyBob:
# isSameString 
takes two strings, compares them lexicographically, but ignores the case of letters.
```
"aaa" === "aaA"
```

PR #1186, opened 2022-06-09
Subject of the function is same string

mohamedLazyBob:
resolve 01-edu/public#1157


Hamzaelkhatri:
you should remove `zipstring/README.md` from this branch. otherwise, everything is good.

mohamedLazyBob:
yes, I've just noticed that. I'll be dropping this PR and creating a new one with just that

PR #1202, opened 2022-06-11
add(subject): print-ascii

Hamzaelkhatri:


subjects/printascii/README.md
```diff
@@ -0,0 +1,27 @@
+## print-ascii
+
+### instructions
+
+Write a program that prints the ASCII value of a character passed as an argument to the program
+- If the argument is not a letter nothing will be printed
```
mohamedLazyBob, line 6: character word is a broad word, it can represent text, numbers, or symbols as long as they are stored in one byte, maybe replace it with a printable character and accept all the range 32<= x < 127 or to an English alphabet letter

Hamzaelkhatri: in this exercise, the student should print ASCII of letters only I write it : 
> If the argument is not a letter nothing will be printed

zainabdnaya (approved):

subjects/printascii/README.md
```diff
@@ -0,0 +1,27 @@
+## print-ascii
+
+### instructions
+
+Write a program that prints the ASCII value of a character passed as an argument to the program
```
davidrobert99, line 5: I would consider changing this part to "Write a program that prints the ASCII value of a letter passed". Also, it is said bellow that the program should print (`'\n'`) at the end of the output. Does that apply to this case?

subjects/printascii/README.md
```diff
@@ -0,0 +1,27 @@
+## print-ascii
+
+### instructions
```
davidrobert99, line 3: Instructions with "I" instead of "i"

subjects/printascii/README.md
Hamzaelkhatri: fixed

MSilva95 (approved):

PR #1203, opened 2022-06-12
chaikin audit fix

uristdwarf:
By the raid specification, there can be no animation unless there are more than 2 points

### DIFF
--- a/subjects/chaikin/audit/README.md
+++ b/subjects/chaikin/audit/README.md
@@ -10,7 +10,7 @@
 
 ###### Can you confirm that the program draws a small circle around the control points in order to identify them?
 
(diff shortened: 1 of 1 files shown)


uristdwarf:
I've also reworded the bit about exiting, leaving it up to the student on how to exit the program. I myself used escape to both clear the screen and exit if the screen is clear, which I don't think is wrong in anyway but some auditor might follow the instructions literally.

davhojt:
Hello @uristdwarf.

Thank you for your pull request. Your corrections about the number of points are very appreciated, thank you.

We will leave it that `escape` is certainly a mechanism which exits the program. But student's may choose to implement other ways to exit.

PR #1209, opened 2022-06-13
add(subject):print-middle

Hamzaelkhatri:


subjects/printmiddle/README.md
```diff
@@ -0,0 +1,22 @@
...
+- If the number of arguments is less than 1, print (`'\n'`)
+- If the number of arguments is **even**, print the middle two arguments with a space (`' '`) between the arguments
+- If the number of arguments is **odd**, print the middle one.
+- Print (`'\n'`) at the end of the output.
+
+### Usage
+
+```console
+$ go run . | cat -e
+$
```
davidrobert99, line 15: I think it is better to remove the '$' after the output, as the function does not return that character.

subjects/printmiddle/README.md
```diff
@@ -0,0 +1,22 @@
...
+$ go run . "Hello World!" | cat -e
+Hello World!$
+$ go run . "Hello World!" "World!" "All" | cat -e
+World!$
```
davidrobert99, line 19: For example, here it would only be 'World!'

davidrobert99 (changes requested):

subjects/printmiddle/README.md
Hamzaelkhatri: yes, you right, but `cat  -e` show if there's any new line after the output.
and the subject says that should add a new line after the output .

davidrobert99: oh yes, sorry for my mistake, you are right, I was running without the cat -e. Let's kept as is.

MSilva95 (approved):

PR #1216, opened 2022-06-13
subject(events): Fix time format mismatch in instructions.

mihkelgering:
Output and tests require this format for duration: {hours}H:{minutes}M:{seconds}S


davhojt:
Thanks @mihkelgering. This is resolved by #1225

PR #1217, opened 2022-06-13
Add devisors subject

zainabdnaya:


Hamzaelkhatri (commented):
it will take a lot of time for large numbers? I guess more than 10 second

Hamzaelkhatri:
it will be a good thing if we limited the numbers to 1000 or 10000.

Hamzaelkhatri (approved):

PR #1221, opened 2022-06-14
DEV-3888-subject(camelTosnackCase): add readme

Hamzaelkhatri:


zainabdnaya (approved):

subjects/cameltosnakecase/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+Here are some rules for you to follow:
+
+- If the string is empty, return an empty string.
+- If the string is not `CamelCase`, return the string unchanged.
+- If the string is `CamelCase`, return the `snake_case` version of the string.
+
+Basic `CamelCase` Capitalization Rules:
+
+- The first letter must be capitalized.
```
MSilva95, line 16: This can be used only if we explain that this program will only take into  account the "upper camel case"

subjects/cameltosnakecase/README.md
```diff
@@ -0,0 +1,54 @@
...
+- The word must not have two capitalized letters together (CamelCAse) not end with a capitalized letter (CamelCasE).
```
MSilva95, line 17: instead of the second "not" leave the "nor" as ti was it looks better.

"(CamelCAse) nor end with a capitalized letter (CamelCasE)."

subjects/cameltosnakecase/README.md
```diff
@@ -0,0 +1,54 @@
+## Camel-to-snake-case 
+
+### Instructions
+Write a function that converts a string from `CamelCase` to `snake_case`.
+
+Camel case is the practice of writing phrases without spaces or punctuation, it indicates the separation of words with a single capitalized letter. Snake case is a style of writing in which each space is replaced by an underscore (_) character.
```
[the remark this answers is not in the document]
Hamzaelkhatri: Yes I get it, I will change the solution and subject to make it for camelCase

subjects/cameltosnakecase/README.md
Hamzaelkhatri: changed

subjects/cameltosnakecase/README.md
```diff
@@ -1,21 +1,21 @@
...
+- If the string is `camelCase`, return the `snake_case` version of the string.
 
-Basic `CamelCase` Capitalization Rules:
+Basic `Lower camelCase` Capitalization Rules:
 
-- The first letter must be capitalized.
-- The word must not have two capitalized letters together (CamelCAse) nor end with a capitalized letter (CamelCasE).
-- No numbers or punctuations are allowed in the word at any place (CamelCase1more).
+- The first letter should be lower case.
+- The word must have only one capitalized letter (camelCase).
```
MSilva95, line 17: what if its 3 words? It will  have more than one capital letter ...

subjects/cameltosnakecase/README.md
```diff
@@ -46,9 +47,10 @@ and the output should be:
 
 ```console
 $ go run .
-Hello_World
-helloWorld
-Camel_To_Snake_Case
+HelloWorld
```
MSilva95, line 50: This is camel as well so it should be changed...

MSilva95 (changes requested):
You need to decide if you will choose to just accept Lower camel or all camelCase and change the readme accordingly ...

subjects/cameltosnakecase/README.md
Hamzaelkhatri: The camelCase contains only two words. I guess if it is more than three words it will be a pascal case. in this subject, I used only lowerCamelCase

subjects/cameltosnakecase/README.md
Hamzaelkhatri: Yes, it is a camel case but an Upper camel case, and I use only the lower camel case in the subject.

subjects/cameltosnakecase/README.md
MSilva95: i don't see anywhere that it only takes 2 words, where did you find this info?

subjects/cameltosnakecase/README.md
MSilva95: then you must e specific in the readme.
Dont use "camelCase" and instead use "lowerCamelCase" otherwise it will create some confusion...

subjects/cameltosnakecase/README.md
Hamzaelkhatri: Yes, you're right. I misunderstood the definition. Resolved.

subjects/cameltosnakecase/README.md
Hamzaelkhatri: Modified.

MSilva95: Still this output is wrong...
The "HelloWorld" is CamelCase so why isn't the output "Hello_World"?

subjects/cameltosnakecase/README.md
```diff
@@ -0,0 +1,56 @@
+## Camel-to-snake-case 
+
+### Instructions
```
MSilva95, line 3: also the readme is still messy...
Use This:

```
### Instructions

Write a function that converts a string from `camelCase` to `snake_case`.

For this exercise you need to know that camelCase has two different writing alternatives that will be accepted:

- lowerCamelCase
- UpperCamelCase

Some rules for writing in camelCase:

- The word does not end on a capitalized letter (CamelCasE).
- No two capitalized letters shall follow directly each other (CamelCAse).
- Numbers or punctuation are not allowed in the word anywhere (camelCase1).

Here are some rules for you to follow:

- If the string is empty, return an empty string.
- If the string is not `camelCase`, return the string unchanged.
- If the string is `camelCase`, return the `snake_case` version of the string.
```

MSilva95 (dismissed):
Take a look at the readme example that i commented and use it or make this one better and less confusing

subjects/cameltosnakecase/README.md
Hamzaelkhatri: @MSilva95 I'm sure that the output, in this case, is the right one.
In the subject, I use only the Lower camel case. 
> Lower camel case is the practice of writing phrases without spaces or punctuation; it indicates the separation of one word or more with a single capitalized letter between each word except the first one. Snake case is a style of writing in which each space is replaced by an underscore (_) character.
 
That is why the output of this case is `HelloWorld ` and not Hello_Word

MSilva95: It is not specified anywhere that only lower camel is to be considered, actually it is asked for the student to
"Write a function that converts a string from camelCase to snake_case."

In my interpretation you are asking for any kind of camel, so that one should be a wrong output.
Also in my opinion it is a better exercise if it accepts both but that is up to you to decide how do you want this exercise to work.
In case you decide to have just the lowerCamelCase, please don't use this as it is misleading:
```
Write a function that converts a string from camelCase to snake_case.
- If the string is not `camelCase`, return the string unchanged.
- If the string is `camelCase`, return the `snake_case` version of the string.
```

Also there are some rule not mentioned in your readme like:

- The word does not end on a capitalized letter (CamelCasE).
- No two capitalized letters shall follow directly each other (CamelCAse).

subjects/cameltosnakecase/README.md
Hamzaelkhatri: Thank you for your suggestion. I change the description to this one.

subjects/cameltosnakecase/README.md
Hamzaelkhatri: Good point, I change the solution and the readme to manage the cases of Lower and Upper camel cases.

jotapero (approved):
Looks Good :alien:

subjects/cameltosnakecase/README.md
```diff
@@ -0,0 +1,64 @@
...
+	"piscine"
+)
+
+func main() {
+	fmt.Println(piscine.CamelToSnakeCase("HelloWorld"))
+	fmt.Println(piscine.CamelToSnakeCase("helloWorld"))
+	fmt.Println(piscine.CamelToSnakeCase("camelCase"))
+	fmt.Println(piscine.CamelToSnakeCase("CAMELtoSnackCASE"))
+	fmt.Println(piscine.CamelToSnakeCase("camelToSnakeCase"))
+	fmt.Println(piscine.CamelToSnakeCase("132322"))
```
tiagocollot, line 49: Instead of using a giving as an input a string only with numbers, I would suggest for example to give a string of words with a number at the end in order to match this condition: "Numbers or punctuation are not allowed in the word anywhere (camelCase1)."

tiagocollot (changes requested):

subjects/cameltosnakecase/README.md
jotapero: fixed

subjects/cameltosnakecase/README.md
```diff
@@ -0,0 +1,64 @@
...
+And its output:
+
+```console
+$ go run .
+Hello_World
+hello_World
+camel_Case
+CAMELtoSnackCASE
+camel_To_Snake_Case
+132322
```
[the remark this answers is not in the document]
jotapero: fixed

tiagocollot (approved):
LGTM! ⚡

PR #1224, opened 2022-06-14
CountStars subject

zainabdnaya:


Hamzaelkhatri (approved):

subjects/countstars/README.md
```diff
@@ -0,0 +1,31 @@
+## CountStars
+
+### Instructions
+Write a function named `CountStar` that makes u fall asleep by counting stars, it takes an integer in the parameter
```
davidrobert99, line 4: misspelled "makes u" -> "makes you"

subjects/countstars/README.md
```diff
@@ -0,0 +1,31 @@
...
+	fmt.Println(CountStars(4))
+	fmt.Println(CountStars(-1))
+}
+```
+
+```console
+$ go run . | cat -e
+1 star...2 star...3 star...4 star...5$
+1 star...2 star...3 star...4$
+No star$
```
davidrobert99, line 30: Shouldn't the last number also have the "start"? 

For example, CountStars(5) shouldn't be "1 star...2 star...3 star...4 star...5 star" instead of "1 star...2 star...3 star...4 star...5".

One improvement could be if the number of starts is bigger than 1, using starts instead of start. Using the same example of CountStars(5), it would be "1 star...2 stars...3 stars...4 stars...5 stars".

Also, the "$" should be removed as it suggests that the function should return it.

davidrobert99 (changes requested):

subjects/countstars/README.md
```diff
@@ -0,0 +1,33 @@
...
+	fmt.Println(CountStars(-1))
+	fmt.Println(CountStars(1))
+}
+```
+
+```console
+$ go run . 
+1 star...2 stars...3 stars...4 stars...5 stars
+1 star...2 stars...3 stars...4 stars
+No star
```
MSilva95, line 31: No stars

subjects/countstars/README.md
```diff
@@ -0,0 +1,33 @@
+## CountStars
+
+### Instructions
+Write a function named `CountStar` that makes you fall asleep by counting stars, it takes an integer in the parameter
+- If the number is negative or equal to 0, return "`No star`"
```
MSilva95, line 5: yours solution returns No Stars in plural

subjects/countstars/README.md
```diff
@@ -0,0 +1,33 @@
+## CountStars
+
+### Instructions
+Write a function named `CountStar` that makes you fall asleep by counting stars, it takes an integer in the parameter
```
MSilva95, line 4: Try this instead:

`Write a function called CountStars that counts the stars up to a number given as an argument.`

MSilva95 (changes requested):

subjects/countstars/README.md
zainabdnaya: Done

subjects/countstars/README.md
zainabdnaya: Done

subjects/countstars/README.md
```diff
@@ -0,0 +1,43 @@
...
+```go
+func CountStars(num int) string {
+
+}
+```
+### Usage
+
+Here is a possible program to test your function :
+
+```go
```
MSilva95, line 21: Fix this example.
The imports are wrong, its missing the package ...

subjects/countstars/README.md
```diff
@@ -0,0 +1,43 @@
+## CountStars
+
+### Instructions
+Write a function called CountStars that counts the stars up to a number given as an argument.
```
MSilva95, line 4: highlight the function name, don't just copy paste my suggestions, follow the structure you used before

subjects/countstars/README.md
```diff
@@ -0,0 +1,43 @@
...
+- If the number is negative or equal to 0, return "`No stars`"
+- No need to manage overflow.
+
+The Library strconv is allowed.
```
MSilva95, line 8: remove this, this is never added in the readme it is in the allowed func.
If i ask for a comment on the allowed funcs you should just add a comment here on github. 
Also this package is not needed, they can create their own Itoa func as they have already done it in the piscine.

MSilva95 (changes requested):

subjects/countstars/README.md
zainabdnaya: Done

subjects/countstars/README.md
zainabdnaya: Done

subjects/countstars/README.md
MSilva95: this could be a level 3 or 4 exercise, if they use the strconv its a level 2 lets not make it that easy

MSilva95 (approved):

PR #1230, opened 2022-06-15
subject(swapName):add readme

Hamzaelkhatri:


zainabdnaya (approved):
good

subjects/swapname/README.md
```diff
@@ -0,0 +1,26 @@
...
+$ go run . | cat -e
+$ go run . "Hello World" "Hello World" | cat -e
+$ go run . "Hello World" | cat -e
+World Hello$
+$ go run . "He2 $%hello" | cat -e
+Error$
+$ go run . "   Hello World" | cat -e
+World Hello$
+$ go run . "Hello World   " | cat -e
+Error$
```
davidrobert99, line 23: Shouldn't the function also ignore the spaces after the second name?

subjects/swapname/README.md
```diff
@@ -0,0 +1,26 @@
+## swap-name
+
+### Instructions
+
+Write a program that takes two words in the first argument and swaps them.
+- If the number of arguments is not 2, nothing should be printed.
+- If the first argument is not a string, the program should print `"Error\n"`.
+- Skip the spaces in the first of the two words.
```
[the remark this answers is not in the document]
Hamzaelkhatri: yes only in the first name, I add more detail to make it understandable.

subjects/swapname/README.md
Hamzaelkhatri: No, just for the first name, but I will add for the second name .

subjects/swapname/README.md
```diff
@@ -0,0 +1,26 @@
+## swap-name
+
+### Instructions
+
+Write a program that takes two words in the first argument and swaps them.
+- If the number of arguments is not 2, nothing should be printed.
+- If the first argument is not a string, the program should print `"Error\n"`.
```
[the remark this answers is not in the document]
Hamzaelkhatri: I add this line to make it more understandable.
> "The first argument should contain only alphabetic characters and spaces, if not, print `"Error\n"` "

subjects/swapname/README.md
```diff
@@ -0,0 +1,26 @@
...
+- If the number of arguments is not 2, nothing should be printed.
+- The first argument should contain only alphabetic characters and spaces and only two word, if not, print `"Error\n"`
+- Print the result with a new line at the end (`'\n'`).
+- Trim space in the beginning and end of the string
+
+### Usage
+
+```console
+$ go run . | cat -e
+$ go run . "Hello World" "Hello World" | cat -e
```
MSilva95, line 15: We should use names as examples since the exercise is named "swap name"

subjects/swapname/README.md
```diff
@@ -0,0 +1,26 @@
+## swap-name
+
+### Instructions
+
+Write a program that takes two words in the first argument and swaps them.
```
MSilva95, line 5: Try this instead:
```
Write a program that takes a first and last name as one argument and swaps them.
- If there is more than one argument, nothing should be printed.
- The argument should contain only alphabetic characters and spaces and only two words, if not, you have to print `"Error\n"`
- The result must be followed with a new line at the end (`'\n'`).
- Trim any extra space in the beginning or end of the string.

File: subjects/swapname/README.md [hunk 6]
```diff
@@ -0,0 +1,26 @@
+## swap-name
+
+### Instructions
+
+Write a program that takes a first and last name as one argument and swaps them.
+- If there is more than one argument, nothing should be printed.
+- The argument should contain only alphabetic characters and spaces and only two words, if not, you have to print `"Error\n"`
+- The result must be followed with a new line at the end (`'\n'`).
+- Trim any extra space in the beginning or end of the string.
```
MSilva95, line 9: Leave it like so :

`Trim any extra spaces in the string.`

Because you also trim the extra spaces in the middle of the arg.
Also add a test for it.

MSilva95 (changes requested):

MSilva95 (approved):

PR #1261, opened 2022-06-20
subject(Add-if-positive):add readme

Hamzaelkhatri:


zainabdnaya (approved):

subjects/addifpositive/README.md
```diff
@@ -0,0 +1,39 @@
...
+package main
+
+import "fmt"
+
+func main() {
+    fmt.Println(AddIfPositive(1, 2))
+    fmt.Println(AddIfPositive(1, -2))
+    fmt.Println(AddIfPositive(-1, 2))
+    fmt.Println(AddIfPositive(-1, -2))
+    fmt.Println(AddIfPositive(10,20))
```
davidrobert99, line 26: I think it is not clear what is supposed to happen when you add the number 0 with another number. Consider adding where in the example.

subjects/addifpositive/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+### Expected function 
+```go
+func AddIfPositive(a int, b int) int {
+    // your code here
+}
+```
+
+### Usage
+
```
davidrobert99, line 15: Missing "Here is a possible program to test your function:"

davidrobert99 (changes requested):

subjects/addifpositive/README.md
Hamzaelkhatri: the subject says that if both numbers are positive it should return the sum of both numbers,
so even if one of them equals 0 or both of them are 0.

davidrobert99: From what I understand, the number Zero is neither positive nor negative. However, Zero is a positive integer since it doesn't carry any negative sign. So it might do a bit of confusion because of the term number and integer.

davidrobert99 (approved):

PR #1263, opened 2022-06-20
unzipstring subject

zainabdnaya:


subjects/unzipstring/README.md
```diff
@@ -0,0 +1,61 @@
...
+$ go run .
+$HHHHHaaaaaaaa
+$pppiiiiiiWWW
+$OOuuuuuHH
+$ff
+$Invalid Input
+$Invalid Input
+$Invalid Input
+$Invalid Input
+
```
Hamzaelkhatri, line 60: delete this empty line

subjects/unzipstring/README.md
```diff
@@ -0,0 +1,61 @@
...
+func main() {
+    fmt.Println(piscine.Unzipstring("6H8a"))
+    fmt.Println(piscine.Unzipstring("3p6i1W"))
+    fmt.Println(piscine.Unzipstring("2O5u2H0e"))
+    fmt.Println(piscine.Unzipstring("2f"))
+    fmt.Println(piscine.Unzipstring("2a 6p8f"))
+    fmt.Println(piscine.Unzipstring("2p4;7g"))
+    fmt.Println(piscine.Unzipstring("2t4dD"))
+    fmt.Println(piscine.Unzipstring("82t4D"))
+
```
Hamzaelkhatri, line 44: delete this empty line

subjects/unzipstring/README.md
```diff
@@ -0,0 +1,61 @@
+## Unzipstring
+
+
+### Instructions
+
+Unzipstring is a simple function that takes a string in of form of a number and an alphabet like (2a) and returns the original string (aa).
```
Hamzaelkhatri, line 6: you can change this line with this : 
> write a function named `Unz...` that takes a string ...
and  I guess there is no need to add an example in the instructions part.

Hamzaelkhatri (changes requested):

subjects/unzipstring/README.md
```diff
@@ -3,7 +3,7 @@
 
 ### Instructions
 
-Unzipstring is a simple function that takes a string in of form of a number and an alphabet like (2a) and returns the original string (aa).
+write a function named Unzipstring that takes a string iN,m form of a number and an alphabet like (2a) and returns the original string (aa).
```
Hamzaelkhatri, line 6: add `` to name of function 
> `Unzipstring(string)`

subjects/unzipstring/README.md
```diff
@@ -0,0 +1,57 @@
+## Unzipstring
+
+### Instructions
+
+write a function named `Unzipstring` that takes a string in form of a number and printable character that is not a number and returns the original string.
```
MSilva95, line 5: This readme is really confusing.
Find a way to explain all correctly, and give an example of the expected format.

MSilva95 (changes requested):

subjects/unzipstring/README.md
```diff
@@ -0,0 +1,54 @@
+## Unzipstring
+
+### Instructions
+
+write a function named `Unzipstring` that takes a string in form of a number and an alphabet (like 3w or 2m3e) then returns a new string (3w ==> www and 2m3e ==> mmeee).
```
MSilva95, line 5: Try this...

Write a function called `Unzipstring` that takes a string that will be a kind of code and your function will have to decrypt it and return a new string with the output.

It works as follows:

The string will be formed by a number followed by a letter, and the purpose is to print this letter the number of times that is requested by the number that precedes it.
```
"3w" ==> www
"2m3e" ==> mmee
```
- The number before the letter must be between 0 and 9.
- The Letter after each number must be between `a and z` or `A and Z`.
- You cannot have two numbers or two letters in a row.
- If the input string does not respect the format, return `Invalid Input`.

MSilva95 (changes requested):

subjects/unzipstring/README.md
```diff
@@ -0,0 +1,59 @@
+## Unzipstring
+
+### Instructions
+
+Write a function called Unzipstring that takes a string that will be a kind of code, and your function will have to decrypt it and return a new string with the output.
```
MSilva95, line 5: Where are the highlights?
Don't just copy paste my comments because git removes the ` from the words
Missing highlight the func name and lines 12, 13 and 15

MSilva95 (changes requested):

MSilva95 (changes requested):

subjects/unzipstring/README.md
```diff
@@ -0,0 +1,59 @@
...
+package main
+
+import "fmt"
+
+func main() {
+    fmt.Println(Unzipstring("2f"))
+    fmt.Println(Unzipstring("2O5u2H0e"))
+    fmt.Println(Unzipstring("3p6i1W"))
+    fmt.Println(Unzipstring("6H8a"))
+    fmt.Println(Unzipstring("2p4;7g"))
```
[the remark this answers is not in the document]
zainabdnaya: the condition of being an alphabet doesn't make sense so I delete it.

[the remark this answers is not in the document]
MSilva95: Why? I think its a good idea, it makes the exercise a bit more challenging so why not having the condition?
Please put it back and fix the solution...

MSilva95 (changes requested):

MSilva95 (approved):

PR #1282, opened 2022-06-22
1140 ping pong

eslopfer:
Add readme file with the instructions

### DIFF
--- a/subjects/pingpong/README.md
+++ b/subjects/pingpong/README.md
@@ -5,8 +5,8 @@
 Write a program that checks if all numbers up to a given number are odd or even.
 The numbers will be between 0 and 100.
 
-- If the number is odd then it prints the number and `ping` as in the example below.
-- If the number is even then it prints the number and `pong` in the same way.
+- If the number is odd then it prints the number and `pong` as in the example below.
+- If the number is even then it prints the number and `ping` in the same way.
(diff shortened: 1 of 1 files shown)


subjects/pingpong/README.md
```diff
@@ -0,0 +1,22 @@
+### Instructions
```
Hamzaelkhatri, line 1: Add name of exercises 
> ## ping-pong

eslopfer: Thank you

Hamzaelkhatri (approved):

subjects/pingpong/README.md
```diff
@@ -0,0 +1,24 @@
+## pingpong
+
+### Instructions
+
+Write a program that checks if all numbers up to a given number are odd or even.
+The numbers will be between 0 and 100.
+
+- If the number is odd then it prints the number and `ping` as in the example below.
```
MSilva95, line 8: In the usage you are saying that the number 2 for example is odd when 2 is an even number ...
This is wrong here you should be saying that 
"If the number is even then it prints the number and `ping` as in the example below."

eslopfer: I'll correct that. Thank you.

MSilva95 (approved):

PR #1292, opened 2022-06-24
1277 Addfront

Hamzaelkhatri:


subjects/skiptheline/README.md
```diff
@@ -1,4 +1,4 @@
-## skip-the-line
+## AddFront
```
eslopfer, line 1: Subject name should be all in lower case to follow convention.

subjects/skiptheline/README.md
```diff
@@ -0,0 +1,38 @@
...
+    fmt.Println(AddFront("Hello", []string{}))
+}
+```
+
+and the output should be:
+
+```console
+$ go run .
+[Hello world]
+[Hello world !]
```
eslopfer, line 36: When I run this program for this example i get this:
```[Hello  world !]``` with an extra space in the middle, but it is difficult to tell by looking at it, wouldn't it be better to compare the length of the slices instead of printing them?

**Review by eslopfer — CHANGES_REQUESTED** [review]

**Review by eslopfer — APPROVED** [review]
Looks good to me!

File: subjects/skiptheline/README.md [hunk 3]
```diff
@@ -0,0 +1,38 @@
+## addfront
+
+### Instructions
+
+Write a function that receives a string and a slice of strings. Return a new slice with the given string prepended.
```
MSilva95, line 5: I think it looks better like so:

```
Write a function that takes a string and a slice of strings, this function will return a new slice of sting with the given string prepended

File: subjects/skiptheline/README.md [hunk 4]
```diff
@@ -0,0 +1,38 @@
+## addfront
```
MSilva95, line 1: The naming of the folder is different. Change it to addfront

subjects/skiptheline/README.md
MSilva95, line 5: We should add this as well:
```
If the given sting is empty you only need to return the given slice

MSilva95 (changes requested):

MSilva95 (approved):

PR #1294, opened 2022-06-28
DEV-3034 - DEV-3034-pimp-my-style-fifo-lifo-typo

eslopfer:
Fix semantics from 'First' to 'Last'


davhojt (approved):
Good spot @eslopfer, thank you for noticing.

Note: This shouldn't really require three commits for changing one word.

PR #1298, opened 2022-06-29
docs: add readme for printevenarguments subject

eslopfer:
Add readme for printevenarguments subject.

### DIFF
--- a/subjects/printevenarguments/README.md
+++ b/subjects/printevenarguments/README.md
@@ -9,7 +9,9 @@ Write a program that receives some arguments from the command line and prints th
 ```console
 $ go run . first second third | cat -e
 second$
-$ go run . a b c d| cat -e
+$ go run . a b c d | cat -e
 b$
 d$
+$ go run . a | cat - e
+$
 ```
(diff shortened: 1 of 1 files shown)


Hamzaelkhatri (approved):
LGTM

subjects/printevenarguments/README.md
```diff
@@ -0,0 +1,15 @@
+## printevenarguments
+
+### Instructions
+
+Write a program that receives some arguments from the command line and prints the ones with an even index followed by a new line. If the number of even arguments is 0, print a new line.
```
MSilva95, line 5: what if the arg is 1, not even but not 0 as well?
```
$ go run . a | cat -e
```

MSilva95 (changes requested):

subjects/printevenarguments/README.md
eslopfer: I guess that's in the description already? That would mean the number of even arguments is 0. I'll add one example of this to the readme. Thanks for your review.

PR #1301, opened 2022-06-30
DEV-3881-docs: Add readme for sumthemall subject

eslopfer:
Add readme for sumthemall subject

### DIFF
--- a/subjects/sumthemall/README.md
+++ b/subjects/sumthemall/README.md
@@ -2,7 +2,7 @@
 
 ### Instructions
 
-Create a program that receives a number of arguments representing whole numbers and displays the sum of the total. If there are not enough arguments or there is an overflow display 0 followed by a new line. If the argument is not a number display 0 followed by a new line.
+Create a program that receives a number of arguments representing whole numbers and displays the sum of the total. If there are not enough arguments or there is an overflow display 0 followed by a new line. If one of the arguments is not a number display 0 followed by a new line.
 
 ### Usage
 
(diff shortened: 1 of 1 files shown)


Hamzaelkhatri (commented):
can you add, if the argument is not a number display `0` or `"Error\n"`

eslopfer:
> can you add, if the argument is not a number display `0` or `"Error\n"`

Wouldn't this be one of the unrealistic scenarios, since we are telling the student that the program is going to be tested with numbers, and we are the ones writing the tests?

Hamzaelkhatri (approved):

subjects/sumthemall/README.md
```diff
@@ -0,0 +1,16 @@
+## sumthemall
+
+### Instructions
+
+Create a program that receives a number of arguments representing whole numbers and displays the sum of the total. If there are not enough arguments or there is an overflow display 0 followed by a new line. If one of the arguments is not a number display 0 followed by a new line.
```
MSilva95, line 5: Do you think its good like this?

```
Create a program that takes integers as arguments and displays the sum total of all of them. If it is not a number, there are not enough arguments, or if there is an overflow, print `0` followed by a newline.
```

MSilva95 (changes requested):

subjects/sumthemall/README.md
tiagocollot: It is fixed.

jotapero (approved):
LGTM :alien:

PR #1315, opened 2022-07-08
DEV-3886-1270 multorsum

eslopfer:
Write readme for multorsum.

### DIFF
--- a/subjects/multorsum/README.md
+++ b/subjects/multorsum/README.md
@@ -7,7 +7,7 @@ Write a function that receives a slice of int and and int representing the initi
 ### Expected function
 
 ```go
-func MultOrSum(ints []int, init int) {
+func MultOrSum(ints []int, init int) int {
 }
 ```
 
--- a/subjects/multorsum/README.md
+++ b/subjects/multorsum/README.md
@@ -2,7 +2,7 @@
 
 ### Instructions
 
-Write a function that receives a slice of int and and int representing the initial value. If the int is odd it should multiply the accumulated value. If it is even, it should be added instead. Return the result. If the slice is empty return 0.
+Write a function that receives a slice of int and an int representing the initial value. If the int is odd it should multiply the accumulated value. If it is even, it should be added instead. Return the result. If the slice is empty return 0.
 
 ### Expected function
 
--- a/subjects/multorsum/README.md
+++ b/subjects/multorsum/README.md
@@ -2,12 +2,12 @@
 
 ### Instructions
 
-Write a function that receives a slice of `int` and an `int` representing the initial value as arguments. 
+Write a function that receives a slice of `int` and an `int` representing the initial value as arguments.
 
 The function should go through the slice and for each `int` check the following restrictions:
 
-- If the `int` is odd, multiply it by the init value. Return the accumulated value after traversing the entire slice. 
-- If the `int` is even, add it to the init value. Return the accumulated value after traversing the entire slice. 
+- If the `int` is odd, multiply it by the init value. Return the accumulated value after traversing the entire slice.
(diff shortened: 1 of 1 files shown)


subjects/multorsum/README.md
```diff
@@ -0,0 +1,35 @@
+## multorsum
+
+### Instructions
+
+Write a function that receives a slice of int and and int representing the initial value. If the int is odd it should multiply the accumulated value. If it is even, it should be added instead. Return the result. If the slice is empty return 0.
```
Hamzaelkhatri, line 5: > If the slice is empty return 0.

The result of your solution `MultOrSum([]int{}, -100)` ==> `-100` not `0`

Hamzaelkhatri (changes requested):

subjects/multorsum/README.md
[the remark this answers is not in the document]
eslopfer: What I meant to say is `a slice of int and an int`, I'll fix the typo and keep the article if that's ok for you. Thanks for your review.

eslopfer: Good catch thanks. I'll fix it.

Hamzaelkhatri (approved):

subjects/multorsum/README.md
```diff
@@ -0,0 +1,35 @@
+## multorsum
+
+### Instructions
+
+Write a function that receives a slice of int and an int representing the initial value. If the int is odd it should multiply the accumulated value. If it is even, it should be added instead. Return the result. If the slice is empty return 0.
```
MSilva95, line 5: Is this the same as you said? Im a bit confused with the readme...

Write a function that receives two arguments, a slice of int and an int representing the initial value.

- If the initial value is odd it should multiply the accumulated value of the slice.
- If it is even, it should be added instead.
- If the slice is empty return 0.

subjects/multorsum/README.md
```diff
@@ -0,0 +1,35 @@
...
+	fmt.Println(MultOrSum([]int{1, 2, 3, 4}, 0 ))
+	fmt.Println(MultOrSum([]int{1, -2, 3, 4}, 0 ))
+}
+```
+
+And its output :
+
+```console
+$ go run . | cat -e
+19$
```
MSilva95, line 32: im not sure i understood well but isn't this supposed to be 30?

I dont understand this and the negative values can you explain a bit more?

MSilva95 (changes requested):

subjects/multorsum/README.md
eslopfer: Thanks for pointing this out, I'll be more specific regarding the initial value. The numbers that define whether is going to be multiplied or added are the ones from the slice.

subjects/multorsum/README.md
eslopfer: I explained in the previous comment. Will add better description to readme.

jotapero (approved):
Looks Good :alien:

subjects/multorsum/README.md
```diff
@@ -0,0 +1,49 @@
+## multorsum
+
+### Instructions
+
+Write a function that receives a slice of `int` and an `int` representing the initial value as arguments. 
+
+You should traverse the slice and for each `int` check the following restrictions:
```
tiagocollot, line 7: I would suggest to rephrase this  as something like: "The function should go through the slice......restrictions:"

tiagocollot (changes requested):

subjects/multorsum/README.md
jotapero: Agreed. Fixed

PR #1328, opened 2022-07-15
docs: Correct description

eslopfer:
Specify that what the function should take is a pointer to the head node of the list, not to the list.

### DIFF
--- a/subjects/listat/README.md
+++ b/subjects/listat/README.md
@@ -2,7 +2,7 @@
 
 ### Instructions
 
-Write a function `ListAt` that takes a pointer to the head of the list `l` and an `int pos` as parameters. This function should return the `NodeL` in the position `pos` of the linked list `l`.
+Write a function `ListAt` that takes a pointer to the head of the list `l` and an `int pos` as parameters. This function should return the pointer to the `NodeL` in the position `pos` of the linked list `l`.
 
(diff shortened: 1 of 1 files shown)


subjects/listat/README.md
```diff
@@ -2,7 +2,7 @@
 
 ### Instructions
 
-Write a function `ListAt` that takes a pointer to the list `l` and an `int pos` as parameters. This function should return the `NodeL` in the position `pos` of the linked list `l`.
+Write a function `ListAt` that takes a pointer to the head of the list `l` and an `int pos` as parameters. This function should return the `NodeL` in the position `pos` of the linked list `l`.
```
davhojt, line 5: This part still remains a contradiction:
> This function should return the `NodeL` in the position `pos` of the linked list `l`

when compared with the return type of the function.
https://github.com/01-edu/public/blob/849cdbf71f1391b9dc23b9be1492ee37c9624ea8/subjects/listat/README.md?plain=1#L18

davhojt (changes requested):

subjects/listat/README.md
eslopfer: Thanks. I'll address that.

davhojt (approved):

PR #1330, opened 2022-07-15
fix(js): replace base64 url trick by temp file

kigiri:
this fix some tests with new version of nodejs


kigiri:
Self approved
image

PR #1332, opened 2022-07-17
docs(go quest 11): correct grammar of NodeI exercises

davhojt:
There are still some improvements to make - but we will make those after the current round of in progress selection piscines


eslopfer (approved):
Looks good to me.

PR #1347, opened 2022-07-27
DEV-3177-new-go-exercise-fromto

Hamzaelkhatri:


subjects/sorted_arg/README.md
```diff
@@ -0,0 +1,33 @@
...
+$ go run . "zA1" | cat -e
+F$
+$ go run . "01Talent" "student" | cat -e
+$
+$ go run . "Hello World" | cat -e
+F$
+$ go run .  "a b c" | cat -e
+F$
+```
+
```
tiagocollot: Missing "And its output :"?

tiagocollot (changes requested):

subjects/fromto/README.md
```diff
@@ -0,0 +1,45 @@
...
+```
+
+### Usage
+
+Here is a possible program to test your function :
+
+```go
+package main
+
+import (
```
tiagocollot, line 26: Since is only importing one package, there is no need for "fmt" to be inside the " (........)"

subjects/fromto/README.md
```diff
@@ -0,0 +1,45 @@
+## fromto
+
+### Instructions
+
+- Write a function that takes two `integers` and returns a string representing the range of numbers from the first to the second.
+- The function can return the numbers in any order.
+- The number must be separated by a comma and a space.
+- If any of the arguments is bigger than 99 or less than 0, the function returns `"invalid"` followed by a newline (`'\n'`).
```
tiagocollot, line 8: should be `invalid` instead of `"invalid"`. Should be (`\n`) instead of (`'\n'`).

subjects/fromto/README.md
```diff
@@ -0,0 +1,45 @@
...
+- Add `0` at the first of any number if it is less than 10.  
+- Add a new line (`'\n'`) at the end of the string.
```
tiagocollot, line 10: Should be (`\n`) instead of (`'\n'`).

tiagocollot (changes requested):

subjects/fromto/README.md
Hamzaelkhatri: fixed

subjects/fromto/README.md
Hamzaelkhatri: I was inspired by the old subjects. 
they do the same thing for the string `"` and char `'`
for example this subject of [`Displaya`](https://github.com/01-edu/public/tree/master/subjects/displaya)

subjects/fromto/README.md
Hamzaelkhatri: [Comment bellow](https://github.com/01-edu/public/pull/1347#discussion_r940418422)
> I was inspired by the old subjects. 
they do the same thing for the string `"` and char `'`
for example this subject of [`Displaya`](https://github.com/01-edu/public/tree/master/subjects/displaya)

tiagocollot (approved):

zainabdnaya (approved):

subjects/fromto/README.md
```diff
@@ -0,0 +1,48 @@
+## fromto
+
+### Instructions
+
+Write a function that takes two `integers` and returns a `string` representing the range of numbers from the first to the second.
```
jotapero, line 5: Suggestion: "(..) `string` showing the range of numbers from the first to the second.

subjects/fromto/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+- The function can return the numbers in any order.
```
jotapero, line 7: This line should not exist. Instructions are explicit enough, it prints in ascending or descending depending on order of `first` and `second` arguments

subjects/fromto/README.md
tiagocollot: It is changed and fixed.

subjects/fromto/README.md
tiagocollot: Thank you, it is fixed.

PR #1352, opened 2022-07-28
DEV-3202-new-go-exercicse-be-zero

Hamzaelkhatri:


tiagocollot (approved):
Everything seems to be correct!

subjects/bezero/README.md
```diff
@@ -0,0 +1,40 @@
+## be-zero
```
tiagocollot, line 1: The title should be "bezero"

tiagocollot (changes requested):

subjects/bezero/README.md
Hamzaelkhatri: I guess not necessary. 
I just separate two word with '-' like kebab-case.
The title could be Bezero , bezero or be-zero

Hamzaelkhatri: Fixed

tiagocollot (approved):

PR #1357, opened 2022-07-29
printstr subject: fix example output

rcatini:
In the example the printed string has no newline at the end: hence the shell prompt should be displayed immediately after the exclamation point (sh/bash default). An alternative modification of the example would consist in printing a percentage symbol after the bang, and the prompt on the new line (zsh behavior), like it was done in https://github.com/01-edu/public/tree/32eddefd0923a53fadb49750323f1bbb6973d39c/subjects/printstr

### DIFF
--- a/subjects/printstr/README.md
+++ b/subjects/printstr/README.md
@@ -30,8 +30,7 @@ And its output :
 
 ```console
 $ go run . | cat -e
-Hello World!
-$
+Hello World!$ 
 ```
 
 ### Notions


rcatini:
NB: this pull request reverts the modifications of 12d57a2f64bca67e37182833df39c9c283da2afb (there is was no explication for putting the dollar symbol on a newline)

subjects/printstr/README.md
```diff
@@ -30,8 +30,7 @@ And its output :
 
 ```console
 $ go run . | cat -e
-Hello World!
-$
+Hello World!$
```
davhojt, line 33: Cannot represent the prompt with $ in this case, because it is confusing with `cat -e`.

davhojt: We will soon implement a different character to represent the prompt.

davhojt (commented):
You may replace the zsh style if you would like to @rcatini? You're welcome to submit a new PR with that change.

We will start replacing the prompt icon shortly.

subjects/printstr/README.md
rcatini: I think the current version it's even more confusing, because `cat -e` is always supposed to print a dollar symbol at the end of line.
I'll wait for the choice of a new prompt, and then I will prepare a pull request for the affected subjects (foreach has a similar issue)

PR #1365, opened 2022-08-01
DEV-3189-new-go-exercise-betterhalf

Hamzaelkhatri:


subjects/secondhalf/README.md
```diff
@@ -0,0 +1,41 @@
+## secondhalf
+
+### Instructions
+
+Write a function `SecondHalf()` that receives a slice of `int` and returns another slice of `int` with the last half of the values.
```
jotapero, line 5: suggestion: (...) with the second half of the values.

jotapero (changes requested):

subjects/secondhalf/README.md
tiagocollot: It makes total sense, thank you. It is fixed.

jotapero (approved):
LGTM :alien:

PR #1368, opened 2022-08-02
DEV-3379-new-go-exercise-swap-last

Hamzaelkhatri:


jotapero (approved):
Looks Good :alien:

eslopfer (dismissed):
What should the learner do if the slice contains just two elements? An example or an explicit instruction would be enough, it's your choice which one to implement :slightly_smiling_face:

mikysett (approved):
LGTM

PR #1374, opened 2022-08-04
DEV-3385-new-go-exercise-only-f

Hamzaelkhatri:


subjects/onlyf/README.md
```diff
@@ -0,0 +1,16 @@
+## onlyf
+
+### Instructions
+
+Write a program that displays an `f` character on the standard output. (and nothing else)
```
tiagocollot, line 5: Instead of "Write a program that displays an `f` character on the standard output. (and nothing else)" I would suggest "Write a program that displays an `f` character on the standard output, and nothing else."

tiagocollot (changes requested):

subjects/onlyf/README.md
jotapero: fixed

tiagocollot (approved):
LGMT!

eslopfer (approved):
LGTM! :rocket:

PR #1377, opened 2022-08-04
DEV-3200-new go exercise hashing code

zainabdnaya:


subjects/hashcode/README.md
```diff
@@ -0,0 +1,48 @@
+## hashcode
+
+### Instructions
+
+Write a function called `HashCode()` that takes a `string` as an argument and return a new **hashed** `string`.
```
jotapero, line 5: Fix: (...) and returns a new (...)

jotapero (changes requested):
Also deal with unnecessary files

subjects/hashcode/README.md
tiagocollot: It is fixed.

PR #1384, opened 2022-08-09
DEV-3410-new go exercise print if not

zainabdnaya:


subjects/printifnot/README.MD
```diff
@@ -0,0 +1,45 @@
+## printifnot
+
+### Instructions
+
+Write a function that takes a `string` as an argument and returns the letter `G` if the argument length is less than 3, otherwise returns **Invalid Input\n**.
```
tiagocollot, line 5: - Instructions fix: "...if the argument's length is less than or equal to 3, otherwise return `Invalid Input\n`."
- With equal to 3, I mean that our solution is including  argument's with the length of 3, we need to decide if it is less than 3 exclusively, because if it is less than 3 then our solution is wrong.

subjects/printifnot/README.MD
```diff
@@ -0,0 +1,45 @@
...
+- If it's an empty string return `G` followed by a newline.
```
tiagocollot, line 6: I would suggest "...empty `string` return `G` followed by a newline `\n`. "

subjects/printifnot/README.MD
```diff
@@ -0,0 +1,45 @@
...
+	"fmt"
+
+	"piscine"
+)
+
+func main() {
+	fmt.Print(piscine.PrintIfNot("abcdefz"))
+	fmt.Print(piscine.PrintIfNot("T2d97d1"))
+	fmt.Print(piscine.PrintIfNot(""))
+	fmt.Print(piscine.PrintIfNot("14"))
```
tiagocollot, line 33: I would suggest to add one case where we have a string with 3 characters.

tiagocollot (changes requested):

subjects/printifnot/README.MD
jotapero: fixed

subjects/printifnot/README.MD
jotapero: fixed

tiagocollot (approved):
LGTM! ⚡

eslopfer (approved):
LGTM! :rocket:

PR #1395, opened 2022-08-17
docs(unwrap_or_expect): rewrite the subject

mikysett:
Rewrite the subject completely to give more meaningful learning experience to students.

rust-tests repo needs to be updated too

### DIFF
--- a/subjects/unwrap_or_expect/README.md
+++ b/subjects/unwrap_or_expect/README.md
@@ -57,6 +57,7 @@ fn main() {
 And its output:
 
 ```console
+$ cargo run
 server1.com
 WARNING: check the server
 Not found: server2.com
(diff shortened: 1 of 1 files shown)


subjects/unwrap_or_expect/README.md
```diff
@@ -53,38 +39,27 @@ Here is a program to test your function:
...
 ```console
-[]
-[2, 4, 6]
-("There is a even value in the vector!", [2])
-[2, 4, 6]
-Ok([2, 4, 6])
-[2, 4, 6]
-Err(("There is a even value in the vector!", [2, 6]))
-[2, 6]
+server1.com
```
davhojt, line 60: Some other exercises have this before the output. Should we include that here also?
```
$ cargo run
```

davhojt (approved):
I left one comment about the console missing command, but it is more about consistency with other exercises. @mikysett can investigate and change if required.

subjects/unwrap_or_expect/README.md
mikysett: Yes, I've added it for consistency

PR #1396, opened 2022-08-18
docs(looping): Removing one edge case

mikysett:
Now there is no difference in output for the word "trial" in case the user succeeds on the first one.

rust-test needs to be updated also to reflect the change.


davhojt (approved):

davhojt:
This should be merged at the same time as https://github.com/01-edu/rust-tests/pull/51

PR #1403, opened 2022-08-29
DEV-3504-feat(piscine-go): add README.md for new go exercise descendappendrange

tiagocollot:


jotapero (approved):
missclicked on "Approve" instead of "Request"

subjects/descendappendrange/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+### Instructions
+
+- Write a function that takes an `int` max and an `int` min as parameters. The function should return a slice of `int`s with all the values between the max and min.
+
+- The `Max`must be included, and min must be excluded.
+
+- If max is inferior than or equal to min, a `nil` slice is returned.
+
+- `make`is not allowed for this exercise
```
jotapero, line 11: Should specify make as function : " make() is not allowed ..."

subjects/descendappendrange/README.md
```diff
@@ -0,0 +1,46 @@
+## descendappendrange
+
+### Instructions
+
+- Write a function that takes an `int` max and an `int` min as parameters. The function should return a slice of `int`s with all the values between the max and min.
+
+- The `Max`must be included, and min must be excluded.
```
jotapero, line 7: max instead of Max, variables declared with lowercase. Also both min and max inside "``" or both without.

jotapero (changes requested):

subjects/descendappendrange/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+- If max is inferior than or equal to min, a `nil` slice is returned.
```
jotapero, line 9: Nil slice does not exist. Either have "Empty slice" or return a `nil` value. 
"If max is inferior than or equal to min, (an empty slice)/(a `nil` value) is returned.

jotapero (changes requested):

jotapero (approved):
Everything in order!

jotapero (approved):

subjects/descendappendrange/README.md
```diff
@@ -0,0 +1,46 @@
+## descendappendrange
+
+### Instructions
+
+- Write a function that takes an `int` max and an `int` min as parameters. The function should return a slice of `int`s with all the values between the max and min.
+
+- The `max`must be included, and `min` must be excluded.
```
eslopfer, line 7: Missing space after max

eslopfer (changes requested):

subjects/descendappendrange/README.md
```diff
@@ -0,0 +1,46 @@
...
+  fmt.Println(piscine.DescendAppendRange(5, 10))
+}
+```
+
+And its output:
+
+```console
+$ go run .
+[10 9 8 7 6]
+[]
```
eslopfer, line 44: Shouldn't it return nil?

eslopfer (changes requested):

jotapero (approved):

subjects/descendappendrange/README.md
jotapero: Appears the purpose is to really return an empty array

jotapero (approved):

subjects/descendappendrange/README.md
eslopfer: Your solution returns nil, if you tell the learner to return an empty slice, when testing it it would fail.

jotapero: Isn't `[]` and empty slice?

eslopfer (approved):

eslopfer (approved):
LGTM! :rocket:

PR #1405, opened 2022-08-31
DEV-3505-feat(piscine-go): add README.md for new go exercise rockandroll

tiagocollot:
Created README.md for the new go exercise rockandroll

### DIFF
--- a/subjects/rockandroll/README.md
+++ b/subjects/rockandroll/README.md
@@ -1,4 +1,4 @@
-# rockandroll
+## rockandroll
 
 ### Instructions
 
@@ -23,14 +23,14 @@ Here is a possible program to test your function:
 package main
 
 import (
-	"fmt"
-	"strconv"
+  "fmt"
+  "piscine"
 )
 
 func main() {
-	fmt.Println(RockAndRoll(4))
-	fmt.Println(RockAndRoll(9))
-	fmt.Println(RockAndRoll(6))
+  fmt.Println(piscine.RockAndRoll(4))
+  fmt.Println(piscine.RockAndRoll(9))
+  fmt.Println(piscine.RockAndRoll(6))
 }
 ```
 And its output:
--- a/subjects/rockandroll/README.md
+++ b/subjects/rockandroll/README.md
@@ -7,6 +7,8 @@ Write a function called `RockAndRoll()` that takes an `int` and returns a `strin
 - If the number is divisible by 2, print `rock` followed by a newline `\n`.
 - If the number is divisible by 3, print `roll` followed by a newline `\n`.
 - If the number is divisible by 2 and 3, print `rock and roll` followed by a newline `\n`.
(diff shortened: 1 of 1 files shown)


subjects/rockandroll/README.md
```diff
@@ -0,0 +1,45 @@
+## rockandroll
+
+### Instructions
+
+Write a function called `RockAndRoll` that takes an `int` and returns a `string`.
+
+- If the number is divisible by 2, print `rock` followed by a newline `\n`.
+- If the number is divisible by 3, print `roll` followed by a newline `\n`.
+- If the number is divisible by 2 and 3, print `rock and roll` followed by a newline `\n`.
+
```
jotapero, line 10: Missing two edge cases (maybe three depending on whether we should test for overflows or not). What happens with negative numbers, and numbers that are not divisible by 2 or 3). Function prints the number on these two cases

subjects/rockandroll/README.md
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"fmt"
+	"strconv"
+)
+
+func main() {
+	fmt.Println(RockAndRoll(4))
+	fmt.Println(RockAndRoll(9))
+	fmt.Println(RockAndRoll(6))
```
jotapero, line 33: Maybe add here edge cases or simply mentioning them above is enough. For instance:
fmt.Println(RockAndRoll(-10)) = ? what should it print
fmt.Println(RockAndRoll(11)) = ?

jotapero (changes requested):

subjects/rockandroll/README.md
tiagocollot: I think it is enough that the test already covers these cases There is no need to be present in the subject.

subjects/rockandroll/README.md
tiagocollot: Same logic as the above answer!

tiagocollot: Same logic as the answer above.

subjects/rockandroll/README.md
tiagocollot: I think it is enough, the test already covers these cases. There is no need to be present in the subject.

jotapero (approved):

subjects/rockandroll/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+func main() {
+  fmt.Println(piscine.RockAndRoll(4))
+  fmt.Println(piscine.RockAndRoll(9))
+  fmt.Println(piscine.RockAndRoll(6))
+}
+```
+And its output:
+
+```go
```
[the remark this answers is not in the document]
tiagocollot: Yes it should, it is fixed! Thank you.

eslopfer (approved):
LGTM! :rocket:

eslopfer (approved):

Issue #1419, opened 2022-09-06
ascii-art subject - missing escape characters in audit arguments


gritlab:
Escape characters are missing in the following audit questions

1. 
Current argument: "hello There 1 to 2!"
Expected argument: "hello There 1 to 2`escape character`!"
<img width="1066" alt="Screen Shot 2022-09-06 at 14 18 51" src="https://user-images.githubusercontent.com/109672158/188624328-c2c60f96-69b9-4783-8919-10bd35330b64.png">

2.
Current argument: "\!\\" #$%&'()*+,-./"
Expected argument: "`escape character`\\`escape character`!\\" #$%&'()*+,-./"
<img width="1072" alt="Screen Shot 2022-09-06 at 14 18 34" src="https://user-images.githubusercontent.com/109672158/188624684-19408194-cbbd-4611-b740-2347ea7edd7b.png">

PR #1420, opened 2022-09-07
[GitHub issue #1419] ascii-art audit: Escape exclamation marks

davhojt:
The exclamation marks were not properly escaped, so the actual output did not match the expected result in the audit file.

Here is a small program to test args with node

##### index.js
```js
console.log(process.argv[2])
```

##### commands
```console
$ node index.js "\\!\" #$%&'()*+,-./"
```

```console
$ node index.js "hello There 1 to 2\!"
```


tiagocollot (approved):
LGTM!

PR #1437, opened 2022-09-12
1433 union inconsistent newline requirement between example and test

eslopfer:
Correct usage example to print new line


davhojt (approved):
LGTM.

PR #1445, opened 2022-09-14
1431 traits subject

mikysett:
as per student suggestion, `fat_content` is now described as "decimal fraction" and not as percentage


davhojt (approved):
Link to the GitHub or Jira issue.

PR #1451, opened 2022-09-15
1448 sales subject

mikysett:
Improved the subject as per student feedback and also solved two typos


davhojt (approved):
Missing GitHub or Jira ticket link.

PR #1459, opened 2022-09-16
docs(ascii-art-justify): update audit to keep consistency with subject

nprimo:


davhojt (changes requested):
I think the problem reported in the issue still remains, in that there is a difference between `"EX"` and `"Example"` in the usage outputs.
https://github.com/01-edu/public/blob/2bcd66760a7cc561b1d8577d66ed7b9d5d618619/subjects/ascii-art/justify/audit.md?plain=1#L10
https://github.com/01-edu/public/blob/2bcd66760a7cc561b1d8577d66ed7b9d5d618619/subjects/ascii-art/justify/README.md?plain=1#L27

davhojt (approved):
I checked out some of the other optional projects of ascii-art. They seem to use "EX", but each of those is scoped to that particular project so "Example" is okay here.

PR #1466, opened 2022-09-21
docs(ascii-art-justify): remove double quotes

nprimo:
clarify how parameters should be passed and add escape char to avoid opend double quote

### DIFF
--- a/subjects/ascii-art/justify/audit.md
+++ b/subjects/ascii-art/justify/audit.md
@@ -2,7 +2,7 @@
 
 ###### Has the requirement for the allowed packages been respected? (Reminder for this project: only [standard packages](https://golang.org/pkg/))
 
-##### Try passing as arguments `"something standard --align right"`
+##### Try passing as arguments `banana standard --align right`
 
 ```
 Usage: go run . [STRING] [BANNER] [OPTION]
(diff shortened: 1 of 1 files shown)


subjects/ascii-art/justify/audit.md
```diff
@@ -12,51 +12,51 @@ Example : go run . something standard --align=right
...
 
 ###### Does it display the correct result at the right side?
 
-##### Try passing as arguments `"#$%&" thinkertoy --align=center"`
+##### Try passing as arguments `#$%&\" thinkertoy --align=center`
 
 ###### Does it display the correct result at the center?
 
-##### Try passing as arguments `""23Hello World!" standard --align=left"`
+##### Try passing as arguments `"23Hello World!" standard --align=left`
```
davhojt, line 43: Exclamation marks need to be escaped.

subjects/ascii-art/justify/audit.md
```diff
@@ -12,51 +12,51 @@ Example : go run . something standard --align=right
...
 
 ###### Does it display the correct result at the left side?
 
-##### Try passing as arguments `""HELLO there HOW are YOU?!" thinkertoy --align=justify"`
+##### Try passing as arguments `"HELLO there HOW are YOU?!" thinkertoy --align=justify`
```
davhojt, line 47: Exclamation marks need to be escaped.

davhojt (changes requested):
Some special characters need to be escaped.

You can use `echo` to test that command line arguments will work as expected.

subjects/ascii-art/justify/audit.md
nprimo: I didn't know the `!` was a special character for go!

davhojt: It isn't special for go, it's rather that it is special for command-line arguments. It can be strange even depending on whether it is wrapped in single or double quotes. You can see with `echo`:
```console
bash-3.2$ echo hello!!!!!
bash: !: event not found
```

```console
bash-3.2$ echo "hello!!!"
bash: !": event not found
```

```console
bash-3.2$ echo 'hello!!!!'
hello!!!!
```

davhojt (approved):
Looks good now, cheers @nprimo.

PR #1468, opened 2022-09-21
1460-DEV-3511 feat(iterators): improve consistency and clarity

mikysett:
now the iterator trait is well implemented and tested and collatz() return value in line with idiomatic rust first example formatting is clearer
provided testing main() has also edge cases 0 and 1

I will merge tests and public at the same time to avoid inconsistencies.


davhojt (approved):
LGTM.

PR #1472, opened 2022-09-22
Dev 3510 features highest missing information in expected functions

eslopfer:


subjects/highest/README.md
```diff
@@ -33,9 +41,6 @@ Here is a program to test your function.
 use highest::*;
 
 #[derive(Debug)]
```
mikysett, line 43: You should remove this `derive` directive (and maybe also the extra empty line behind it).

mikysett (approved):
Everything looks fine, I just leaved a comment in the `Usage` part.
There is one `derive` directive left there (the one at line 43) that should be removed.

After that change you can merge.

PR #1486, opened 2022-10-06
feat(shopping_mall): remove not neutral function

mikysett:
also improve readability of output for the example main.rs


eslopfer (approved):
LGTM! :rocket:

PR #1487, opened 2022-10-06
docs(does_it_fit): introduce functions arguments more explicitly

mikysett:


nprimo (approved):
It looks good to me :)

PR #1488, opened 2022-10-06
docs(diamond_creation): improve code example clarity

mikysett:


nprimo (approved):
It looks good and clearer than before 👍

PR #1495, opened 2022-10-10
fix(filler): placing the bonus questions inside a bonus part

MSilva95:


mikysett (approved):
It seems clearer to me, still isn't it part of the guidelines to use `#### General` for bonus questions? In any case for the current issue I think we can proceed

PR #1500, opened 2022-10-11
Update audit.md

sagarishere:
Escaping ! with \! so that terminal does not start special commands and go file receives the command.


sagarishere:
escaping special character ! with use of backslash

nprimo:
@sagarishere, thank you so much for opening this PR. I will close it since it is now not updated with the current version of the audit of this project

PR #1502, opened 2022-10-12
DEV-3237 Data Simple hash refactory to have a more useful and stimulating intro to hash maps

mikysett:
need to be merged at the same time of rust-tests.


eslopfer (approved):
LGTM! :rocket:

PR #1504, opened 2022-10-12
DEV-3470 Docs-Easy-traits Fix prototype error and add pub where necessary

mikysett:
also improve the arguments names

to update at the same time of `rust-tests`


eslopfer (approved):
LGTM! :rocket:

PR #1505, opened 2022-10-12
DEV-3192 Refactor-Tuples-Refs Change id type for more idiomatic rust

mikysett:
need to be merged at the same time of `rust-tests`


eslopfer (approved):
LGTM! :rocket:

PR #1513, opened 2022-10-13
DEV-3198-new-go-exercise-get-digit-len-fix

tiagocollot:
Added README.md for Go exam digitlen.


subjects/digitlen/README.md
```diff
@@ -0,0 +1,46 @@
+## digitlen
+
+### Instructions
+
+Write a function `DigitLen()` that takes two integers as arguments and returns the number of digits in the first `int` by the base of the second `int`.
```
jotapero, line 5: Fix: (instructions are not that clear) "Write a function `DigitLen()` that takes two integers as arguments and returns the times the first `int` can be divided by the second until it reaches zero"

jotapero (changes requested):

subjects/digitlen/README.md
tiagocollot: Thank you José. It is fixed.

jotapero (approved):
LGTM :alien:

subjects/digitlen/README.md
```diff
@@ -0,0 +1,46 @@
+## digitlen
+
+### Instructions
+
+Write a function `DigitLen()` that takes two integers as arguments and returns the times the first `int` can be divided by the second until it reaches zero.
+
+- The second `int` must be between ***2*** and ***36***. If not, the function returns `-1`.
```
eslopfer, line 7: When I ran prettier it complained here.

eslopfer (changes requested):

PR #1514, opened 2022-10-14
Update audit.md

sagarishere:
If one wants the dollars signs as line endings visually, I suggest we provide the right command. ` cat -e `


tcellect (approved):

nprimo:
As with the previous PR, I will close this since it's not relevant to the current status of the project's audit. Thank you in any case for the feedback @sagarishere

PR #1517, opened 2022-10-16
Update audit.md

sagarishere:
escaping special command parameters for bash terminal


davhojt (dismissed):
Thanks @sagarishere

nprimo:
Thank you for pointing this out. We have worked out another solution to be consistent with all OS

PR #1525, opened 2022-10-19
DEV-3378-new-go-exercise-quarterofayear

tiagocollot:


jotapero (approved):
Looks Good :alien:

subjects/quarterofayear/README.md
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"fmt"
+	"piscine"
+)
+
+func main() {
+	fmt.Println(piscine.QuarterOfAYear(2))
+	fmt.Println(piscine.QuarterOfAYear(5))
+	fmt.Println(piscine.QuarterOfAYear(8))
+	fmt.Println(piscine.QuarterOfAYear(11))
```
eslopfer, line 34: It would be nice to have an example of a month that lies in the 'frontier' between quarters, and less examples of similar cases, for instance, 2, 5 and 8, are all the second month of the quarter, it would be enough with only one

eslopfer (changes requested):

PR #1533, opened 2022-10-20
DEV-3213 Borrow-me-the-reference Remove need to use meval crate

mikysett:
need to be merged at same time of `rust_tests`


eslopfer (approved):
LGTM! :rocket:

Issue #1304, opened 2022-06-30
printstr subject


CIRCADES:
printstr
---

The hint suggests output:

$ go run . | cat -e
Hello World!%
$

The "%" is probably a mistake. Also, the Test Panel expects no new line.

davhojt:
Hello @CIRCADES.

The `%` is normally used to represent that there is no new line at the end. This is however not consistent across different shells.

Some shells place the prompt directly after the output. But as `cat -e` used `$` to represent the end of the line, and the prompt is also represented by a `$`, it gets a bit confusing to use it.

https://unix.stackexchange.com/questions/167582/why-zsh-ends-a-line-with-a-highlighted-percent-symbol

ohmyzsh:
<img width="350" alt="image" src="https://user-images.githubusercontent.com/22807573/176935398-876813e6-8070-4011-83fe-7a7ee08cce2c.png">

bash:
<img width="243" alt="image" src="https://user-images.githubusercontent.com/22807573/176935700-9480e03d-769c-4168-8d6c-39efc4592140.png">

csh:
<img width="441" alt="image" src="https://user-images.githubusercontent.com/22807573/176935994-5c6e712a-03ad-4cd9-b088-3a461649d9f5.png">

PR #1542, opened 2022-10-31
DEV-4001 Bring issue from public repo - printstr subject

eslopfer:
There was an extra new line in the description's example.

### DIFF
--- a/subjects/printstr/README.md
+++ b/subjects/printstr/README.md
@@ -29,8 +29,8 @@ func main() {
 And its output :
 
 ```console
-$ go run . | cat -e
-Hello World!$
+go run . | cat -e
+Hello World!
 ```
 
 ### Notions
--- a/subjects/printstr/README.md
+++ b/subjects/printstr/README.md
@@ -30,8 +30,7 @@ And its output :
 
 ```console
 $ go run . | cat -e
-Hello World!
-$
+Hello World!$
 ```
 
 ### Notions


subjects/printstr/README.md
```diff
@@ -30,8 +30,7 @@ And its output :
 
 ```console
 $ go run . | cat -e
-Hello World!
-$
+Hello World!$
```
davhojt, line 33: This is tricky to fix at the moment. This is valid, but also misleading. The `$` represents both the prompt and `cat -e` new-line position. So it is difficult to understand which is being used here. See https://github.com/01-edu/public/issues/1304.

One thing we could try here is to represent the prompt and new-line position with different characters. Perhaps the prompt could be something like `prompt>` or something else appropriate.

davhojt (changes requested):
It's a tough one to get right. At some point we need to make a solution for all the places where this is an issue.

subjects/printstr/README.md
eslopfer: I see what you mean, it makes sense. Wouldn't it be less confusing then, to remove the $ at the beginning of the command, since different prompts represent it in different ways, and its the source of the problem? Should I remove it as well?

We are specifying that it's the output, so wouldn't it be creating less confusion for the students and thereby less issues to solve? Perhaps modify line 29  then to say `And its output to the terminal:`

I think that even if not ideal it would be better. What do you think?

eslopfer: To be more specific, what I proposed is something like this:
`And its output to the the terminal:`
- without new line required:
```console
go run . | cat -e
Hello World!
```
- with new line required: 
```console
go run . | cat -e
Hello World!$
```

davhojt: It's quite common in console type code-blocks, that commands are distinct from output. I guess we don't actually _need_ to show the command though in this case. I think it was used in this case to demonstrate that the function should print a string which is not new-line terminated (using `cat -e`).

```console
bash-3.2$ echo -n "frogs are cool"
frogs are coolbash-3.2$
```

```console
bash-3.2$ echo -n "frogs are cool" | cat -e
frogs are coolbash-3.2$
```

```console
bash-3.2$ echo "frogs are cool" | cat -e
frogs are cool$
bash-3.2$
```

In the long term, the intention is to make the prompt distinct. But in the short term, you can do whatever makes the exercise more clear.

---

You can implement any fix which makes the subject clearer.

davhojt:
@eslopfer PR name needs to be changed.

subjects/printstr/README.md
eslopfer: Done

eslopfer:
> @eslopfer PR name needs to be changed.

I created a Jira issue to change it. :slightly_smiling_face:

davhojt (approved):
LGTM.

subjects/printstr/README.md
eslopfer: I am resolving because the pull request was approved.

PR #1545, opened 2022-11-03
Fix getNotes

lotoussa:
to avoid TypeError: Cannot read properties of undefined (reading '$$eval')

### DIFF
--- a/js/tests/keycodes-symphony-dom_test.mjs
+++ b/js/tests/keycodes-symphony-dom_test.mjs
@@ -1,12 +1,5 @@
 export const tests = []
 
-export const setup = async ({ page }) => ({
-  getNotes: async () =>
-    await page.$$eval('.note', (nodes) => {
-      return nodes.map((note) => note.textContent)
-    }),
-})
-
 const characters = `didyouhandlethekeydowneventcorrectly`
 
 tests.push(async ({ page, eq, ctx }) => {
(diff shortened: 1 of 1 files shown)


lotoussa:
Need to remove the '(' line 37
  const cleared = (getNotesTc.length === 0

js/tests/keycodes-symphony-dom_test.mjs
```diff
@@ -24,15 +20,21 @@ tests.push(async ({ page, eq, ctx }) => {
...
 })
 
 tests.push(async ({ page, eq, ctx }) => {
   // check that all the notes are cleared when Escape key is pressed
   await page.keyboard.down('Escape')
-  const cleared = (await ctx.getNotes()).length === 0
+  const getNotesTc = await page.$$eval('.note', (nodes) => {
+    return nodes.map((note) => note.textContent)
+  })
+  const cleared = (getNotesTc.length === 0
```
nprimo, line 37: as you pointed out, this line is missing a parenthesis

js/tests/keycodes-symphony-dom_test.mjs
```diff
@@ -1,20 +1,16 @@
...
-
 const characters = `didyouhandlethekeydowneventcorrectly`
 
 tests.push(async ({ page, eq, ctx }) => {
   // check that a note is created and matches the right letter when a key is pressed
   for (const [i, character] of characters.split('').entries()) {
     await page.keyboard.down(character)
     const typed = characters.slice(0, i + 1).split('')
-    eq(await ctx.getNotes(), typed)
+    const getNotesTc = await page.$$eval('.note', (nodes) => {
```
nprimo, line 10: is there any special reason why the new function is `getNotesTc` instead of `getNotes`?

nprimo (changes requested):
I have left a couple of inline comments; beside that I would have declared the new `getNotesTc` (or simply `getNotes`) in the main scope to avoid repetition

js/tests/keycodes-symphony-dom_test.mjs
lotoussa: Tc = TextContent (note.TextContent)

lotoussa:
closed, nprimo exported this pr #1559

PR #1549, opened 2022-11-07
DEV-3877-error-in-statement-in-exercise-reduce-in-call-me-maybe

eslopfer:
- The description said 3 functions, but then it asked to make 4.
- The number of examples was also inconsistent


davhojt (approved):
LGTM.

PR #1552, opened 2022-11-07
DEV-3800 new exercises for rust exam: min_and_max

nprimo:
new exercise for rust exam: `min_and_max`

### DIFF
--- a/subjects/min_and_max/README.md
+++ b/subjects/min_and_max/README.md
@@ -17,13 +17,9 @@ Here is a program to test your function
 use min_and_max::min_and_max;
 
 fn main() {
-    let nb_1 = 9;
-    let nb_2 = 4;
-    let nb_3 = 2;
-    let (min, max) = min_and_max(nb_1, nb_2, nb_3);
     println!(
-        "The minimum is {}, the maximum is {}",
-        min, max
+        "Minium and maximum are: ${?}",
+        min_and_max(9, 2, 4)
     );
 }
 ```
@@ -38,8 +34,8 @@ $
 
 ### Notions
 
-- [The Tuple Type](https://doc.rust-lang.org/stable/book/ch03-02-data-types.html?highlight=accessing%20a%20tuple#compound-types)
+- [The Tuple Type](https://doc.rust-lang.org/stable/book/ch03-02-data-types.html)
 
 - [Tuples](https://doc.rust-lang.org/rust-by-example/primitives/tuples.html)
 
-- [Tuple Structs without Named Fields](https://doc.rust-lang.org/stable/book/ch05-01-defining-structs.html?highlight=tuple#using-tuple-structs-without-named-fields-to-create-different-types)
+- [Tuple Structs without Named Fields](https://doc.rust-lang.org/stable/book/ch05-01-defining-structs.html)
--- a/subjects/min_and_max/README.md
+++ b/subjects/min_and_max/README.md
@@ -18,7 +18,7 @@ use min_and_max::min_and_max;
 
 fn main() {
     println!(
-        "Minium and maximum are: ${?}",
+        "Minium and maximum are: ${:?}",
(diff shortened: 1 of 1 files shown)


subjects/min_and_max/README.md
```diff
@@ -0,0 +1,45 @@
...
+Here is a program to test your function
+
+```rust
+use min_and_max::min_and_max;
+
+fn main() {
+    let nb_1 = 9;
+    let nb_2 = 4;
+    let nb_3 = 2;
+    let (min, max) = min_and_max(nb_1, nb_2, nb_3);
```
mikysett, line 23: maybe we can get rid of the 3 variables and write the arguments inline.
I think it's better if our examples stays as minimal as possible, and students can infer information instead of just read it.

subjects/min_and_max/README.md
```diff
@@ -0,0 +1,45 @@
...
+    println!(
+        "The minimum is {}, the maximum is {}",
+        min, max
+    );
```
mikysett, line 27: in that sense also the println can be simplified by printing the tuple with debug printing.
Like so `println!("{:?}", min_and_max(9, 4, 2));`.

subjects/min_and_max/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+```console
+$ cargo run
+The minimum is 2, the maximum is 9
+$
+```
+
+### Notions
+
+- [The Tuple Type](https://doc.rust-lang.org/stable/book/ch03-02-data-types.html?highlight=accessing%20a%20tuple#compound-types)
```
mikysett, line 41: following the link there is an annoing highlight of some text, removing the `hightlight=*` part from the link should solve the issue.
This is valid also for the other links (and actually this issue should be fixed also in the original exercise)

mikysett (changes requested):
All good, just some style notes to improve code examples :)

subjects/min_and_max/README.md
```diff
@@ -0,0 +1,41 @@
...
+### Usage
+
+Here is a program to test your function
+
+```rust
+use min_and_max::min_and_max;
+
+fn main() {
+    println!(
+        "Minium and maximum are: ${?}",
```
mikysett, line 21: `${?}` it's not a valid syntax for rust fmt, should be replaced with `{:?}`.

mikysett (changes requested):
Small issue to solve on the `println` in `main`

subjects/min_and_max/README.md
```diff
@@ -0,0 +1,41 @@
...
+### Usage
+
+Here is a program to test your function
+
+```rust
+use min_and_max::min_and_max;
+
+fn main() {
+    println!(
+        "Minium and maximum are: ${:?}",
```
mikysett, line 21: The `$` is not necessary and should be removed as it can be confused with some mandatory special char

subjects/min_and_max/README.md
```diff
@@ -0,0 +1,41 @@
...
+        min_and_max(9, 2, 4)
+    );
+}
+```
+
+And its output
+
+```console
+$ cargo run
+The minimum is 2, the maximum is 9
```
mikysett, line 31: This line is not updated considering the new main

mikysett (changes requested):
Still some issues in the code example.
As a nice way to avoid this you could copy the code in a `main.rs` in the `exercise_test/src` and then run `cargo run` in that directory.

Then copy/paste both code and result, so you are sure the code you give works as you expected

subjects/min_and_max/README.md
nprimo: True! Plus I found a typo - `Minium` should be `Minimum`

mikysett (commented):
LGTM

mikysett (approved):

PR #1553, opened 2022-11-07
DEV-3803-exam-01-capitalizing-remake-exercise

eslopfer:
- Description  of modify_letter rust exam exercise.

### DIFF
--- a/subjects/modify_letter/README.md
+++ b/subjects/modify_letter/README.md
@@ -1,0 +1,46 @@
+## modify_letter
+
+### Instructions
+
+Create a **function** `remove_letter_sensitive` that returns a string without the letter specified as argument
+
+Create a **function** `remove_letter_insensitive` that returns a string without the letter specified as argument (ignoring case)
+
+Create a **function** `swap_letter_case` that returns a string swapping the case for the chosen letter.
+
+### Expected Functions
+
+```rust
+pub fn remove_letter_sensitive(s: &str, letter: char) -> String {
+}
+
+pub fn remove_letter_insensitive(s: &str, letter: char) -> String {
+}
+
+pub fn swap_letter_case(s: &str, letter: char) -> String {
+}
+```
+
+### Usage
+
+Here is a program to test your functions.
+
+```rust
+use modify_letter::*modify_letter*;
+
+fn main() {
+    println!("{}", remove_letter_sensitive("Joje jis mijssjing", "j"));
+    println!("{}", remove_letter_insensitive("jaillA ais swiaAmmingA", "A"));
+    println!("{}", swap_letter_case("hEllo therE", "e"));
+}
+```
+
+And its output
+
+```console
+$ cargo run
+Joe is missing
+Jill is swimming
+Hello thERe
+$
+```
--- a/subjects/modify_letter/README.md
+++ b/subjects/modify_letter/README.md
@@ -30,7 +30,7 @@ use modify_letter::*modify_letter*;
 
 fn main() {
     println!("{}", remove_letter_sensitive("Joje jis mijssjing", 'j'));
-    println!("{}", remove_letter_insensitive("jaillA ais swiaAmmingA", 'A'));
+    println!("{}", remove_letter_insensitive("JaillA ais swiaAmmingA", 'A'));
     println!("{}", swap_letter_case("hEllo therE", 'e'));
 }
 ```
@@ -41,6 +41,6 @@ And its output
 $ cargo run
 Joe is missing
 Jill is swimming
-Hello thERe
+hello thEre
 $
 ```
--- a/subjects/modify_letter/README.md
+++ b/subjects/modify_letter/README.md
@@ -29,9 +29,9 @@ Here is a program to test your functions.
 use modify_letter::*modify_letter*;
 
 fn main() {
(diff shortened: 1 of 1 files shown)


subjects/modify_letter/README.md
```diff
@@ -0,0 +1,50 @@
+## modify_letter
+
+### Instructions
+
+Create a **function** `remove_letter_sensitive` that returns a string without the letter specified as argument
```
mikysett, line 5: The sentence is missing an ending `.`.

subjects/modify_letter/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+Create a **function** `remove_letter_insensitive` that returns a string without the letter specified as argument (ignoring case)
```
mikysett, line 7: The sentence is missing an ending `.`.

subjects/modify_letter/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+Create a **function** `swap_letter_case` that returns a string swapping the case for the chosen letter.
+
+If there is an empty character in the arguments, the function should return the string as being passed.
```
mikysett, line 11: This scenario and the following one are part of standard cases, I don't think it's worth writing them down explicitly.

1. If letter is `\0` (which actually is not called "empty character" and it is indeed a character, called EOF) it won't be find in the string (because by definition a string can not have an EOF in it) so the result will be the string unchanged.
2. If the string is empty then it's impossible to modify anything so the same empty string is returned.

subjects/modify_letter/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Usage
+
+Here is a program to test your functions.
+
+```rust
+use modify_letter::*modify_letter*;
+
+fn main() {
+    println!("{}", remove_letter_sensitive("Joje jis mijssjing", 'j'));
```
mikysett, line 36: Maybe it can be nice to modify the strings in the example so they won't be the same of the exercise the students already did in the data quest.

mikysett (changes requested):
Some small style improvements and maybe the need to change examples so they are not identical to the ones already present in other exercises

subjects/modify_letter/README.md
eslopfer: Done, Thanks.

subjects/modify_letter/README.md
eslopfer: Done, Thanks.

subjects/modify_letter/README.md
eslopfer: I am new to rust, so forgive me if I am wrong, but I did some research on this.

Although you are not wrong, since a char is a [Unicode scalar value](https://doc.rust-lang.org/std/primitive.char.html), normally the Unicode null(and I called it 'empty character', [because its all 0s](https://news.ycombinator.com/item?id=3906755), and just because its supported in Unicode doesn't mean we have to, as they discuss in that link) is accepted as the closest you can get [in other languages](https://stackoverflow.com/questions/3670505/why-is-there-no-char-empty-like-string-empty) or [this](https://stackoverflow.com/questions/18410234/how-does-one-represent-the-empty-char),  and what I could find about  it in [rust](https://www.folkstalk.com/2022/07/rust-empty-character-literal-with-code-examples.html), with some controversy of course, as everything opinionated.

Empty strings are generally considered [edge cases](https://www.reddit.com/r/leetcode/comments/u244jn/what_do_people_mean_by_edge_cases/), (there is no authority defining what is or not, so the measure is usually consensus), not part of the standard cases, and as such, sometimes they require being handled separately in a specific way. 

Being said that, I think your suggested handling of the edge case makes common sense. Assuming that it should do what is common sense, is making the learner guess, or try and see if it would work. 

I see 2 options at least, both can encourage the habit of thinking about edge cases:
 - Don't include the instructions and encourage the behaviour of 'lets try and see if it works'(this one could increase frustration levels, but on the other hand the tolerance for it).
 - Include the specific instructions of what to do, and create the habit of thinking about how to handle edge cases and ask for more specific instructions if its not clear before starting to write code(which creates the habit of understanding what to do beforehand, to waste less time modifying it afterwards).
 
Should I still remove the instructions? (That would be the easiest to me, but I want to do what's best.)

subjects/modify_letter/README.md
eslopfer: Thanks, I changed them.

subjects/modify_letter/README.md
eslopfer: This comment is to leave some reference of what we discussed regarding the previous:
- The requested changes where pointing to the solution written in the tests repository already handling this edges without requiring an extra case, not about the cases not being 'edges'.
- Including other edge cases would involve re-evaluating the difficulty of the other exercises to make sure they are on the same level, so for now we agreed it would be better to keep complexity low and to not add any extra checks.
- To resolve this conversation I'll remove the mention to  empty string/empty character and remove them also from the solution.

mikysett (commented):
Perfect, very nice :)

mikysett (commented):
I've triggered a re-review since it was still in state "request for changes" on last review.

Also remember to merge at the same time of `rust-tests` if possible.

mikysett (approved):

PR #1558, opened 2022-11-09
docs(smallest): add readme with description of subject

eslopfer:


subjects/smallest/README.md
```diff
@@ -0,0 +1,44 @@
+## smallest
+
+### Instructions
+
+Create a function named `smallest` that gets the smallest positive number in the `HashMap`.
```
mikysett, line 5: Here the fact we are limiting to "positive numbers" seems very confusing to me.
Why the function should return the smallest positive and not the smallest in general?
Also in that case the function name should change to reflect this specificity (something like smallest_positive).

I think the best is to get rid of this "positive" constraint so the function returns the smallest in general (ex. `[10, 22, -10, 4] -> `-10`).
Let me know if you do agree or if you'd prefer to keep this behavior :)

mikysett (changes requested):

subjects/smallest/README.md
eslopfer: I agree. Thanks for the suggestion.

eslopfer: Done.

subjects/smallest/README.md
```diff
@@ -2,7 +2,9 @@
 
 ### Instructions
 
-Create a function named `smallest` that gets the smallest positive number in the `HashMap`.
+Create a function named `smallest` that gets the smallest number in the `HashMap`.
+
+If the `HashMap` is empty, return the maximum `i32`.
```
mikysett, line 7: I don't think students are ever required to use the max value for i32. We can still ask for them to return it but maybe it will be nice to explicitly tell them it is defined as `i32::MAX`

mikysett (changes requested):
let's just clarify a bit how to use `i32::MAX`

subjects/smallest/README.md
eslopfer: There are several ways to get the max i32, it's something common to (as far as I know) every statically typed language, since it has to do with the amount of numbers that fit in 32 bits, and we are specifying those. If at this point of the program(1 year give it or take) students still don't know I think it's time for them to learn how to find that number.

I don't believe we should spoon feed the way of defining it. It could be hardcoded as a literal, it could be calculated as power of 2, it could be gotten by bit-shifting... worst case scenario they will get there by trial and error. If they don't know what is the maximum number you can fit in certain amount of bits it's reckoning time.

mikysett: It will be challenging for them to find this info during the exam but ok, we can keep it like that and rely on their problem solving attitude.

Still there is some misconceptions here that I rather clarify:
At the moment in our curriculum Rust comes after some time spent programming with other languages but this is not fixed in stone, quite the contrary.
The Rust piscine is meant to be accessible for people that never programmed and should have no technical prerequisites (as it goes for the Go Piscine). Ideally the Rust piscine could be given to schools as a standalone product or schools should be confident that switching Go Piscine and Rust Piscine won't create any issue to new students.

mikysett (approved):
I resolved the conversion after commenting it, I report here the comment since I don't want it to block the merge:

It will be challenging for them to find this info [MAX for i32] during the exam but ok, we can keep it like that and rely on their problem solving attitude.

Still there is some misconceptions here that I rather clarify:
At the moment Rust comes after some time spent programming with other languages but this is not fixed in stone, quite the contrary.
The Rust piscine is meant to be accessible for people that never programmed and should have no technical prerequisites (as it goes for the Go Piscine). Ideally the Rust piscine could be given to schools as a standalone product or schools should be confident that switching Go Piscine and Rust Piscine won't create any issue to new students.

eslopfer:
> I resolved the conversion after commenting it, I report here the comment since I don't want it to block the merge:
> 
> It will be challenging for them to find this info [MAX for i32] during the exam but ok, we can keep it like that and rely on their problem solving attitude.
> 
> Still there is some misconceptions here that I rather clarify: At the moment Rust comes after some time spent programming with other languages but this is not fixed in stone, quite the contrary. The Rust piscine is meant to be accessible for people that never programmed and should have no technical prerequisites (as it goes for the Go Piscine). Ideally the Rust piscine could be given to schools as a standalone product or schools should be confident that switching Go Piscine and Rust Piscine won't create any issue to new students.

I understand your point. During the go piscine, I was challenged several times with integer overflows. It wasn't specified anywhere how to find about it. It was annoying at the time, today I am grateful for that. So if the rust piscine ever replaces the go piscine, this even makes a stronger argument, it will make both piscines more similar in content.

PR #1560, opened 2022-11-10
DEV-3801 add new rust exam exercise - matrix_multiplication

nprimo:


subjects/matrix_multiplication/README.md
```diff
@@ -0,0 +1,53 @@
+## matrix_multiplication
+
+### Instructions
+
+- Define a `struct` named `Matrix` as a tuple of 2 tuples. The nested tuple will contain 2 `i32`s.
```
mikysett, line 5: here it will be better style to say "it will contain two" instead of "it will contain 2"

subjects/matrix_multiplication/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+- Create a **function** named `multiply` that receives a Matrix and an i32 and returns the Matrix with each number multiplied by the second argument.
```
mikysett, line 7: Matrix and i32 should be `Matrix` and `i32`

subjects/matrix_multiplication/README.md
```diff
@@ -0,0 +1,53 @@
+## matrix_multiplication
+
+### Instructions
+
+- Define a `struct` named `Matrix` as a tuple of 2 tuples. The nested tuple will contain 2 `i32`s.
+
+- Create a **function** named `multiply` that receives a Matrix and an i32 and returns the Matrix with each number multiplied by the second argument. 
+```rust
+pub fn multiply(m: Matrix, val: i32) -> Matrix {
```
mikysett, line 9: `val` could be `multiplier` or something like that, as soon as it is more informative.
As a general rule it is better to avoid names like `val` which are highly uninformative and may suggest bad practices to students.

subjects/matrix_multiplication/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+> Remember that you are defining a library, so any element that can be called from an external crate must be made public.
+
+### Usage
+
+Here is a possible program to test your function
+
+```rust
+use matrix_multiplication::multiply;
+use matrix_multiplication::Matrix;
```
mikysett, line 23: When having more than one element to be included from an external crate you may use `use matrix_multiplication::*;`.
This is also valid for your `tests` in `rust-tests`

mikysett (changes requested):
couple of typos and style improvements

PR #1568, opened 2022-11-14
DEV-3813-exam-03-roman-numbers-remake-exercise

mikysett:
merge `public` and `rust-tests` at the same time

### DIFF
--- a/subjects/office_worker/README.md
+++ b/subjects/office_worker/README.md
@@ -11,6 +11,8 @@ Create an enum `WorkerRole` which can be `Admin`, `User` or `Guest`.
 
 Implement for both the trait `From<&str>`. For `OfficeWorker` the string will have the format `"name,age,role"`, for `WorkerRole` the format of the string will be the `"role"` name in lower case.
 
+> Invalid inputs won't be tested.
+
 ### Expected Functions and Data Structures
 
 ```rust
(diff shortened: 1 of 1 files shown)


nprimo (commented):
What should happen when the string to parse the new `OfficeWorker` has more or fewer fields than expected?
What should happen when the string to parse the `WorkerRole` is not correctly formatted - e.g. does not match the cases described in the subject?

nprimo (changes requested):
The subject should specify what happens when there are some wrong format inputs for the `WorkerRoler` fields

mikysett:
I've added a note specifying that invalid inputs won't be tested

subjects/office_worker/README.md
```diff
@@ -0,0 +1,58 @@
+## office_worker
+
+### Instructions
+
+Create a structure `OfficeWorker` having the following public fields:
```
nprimo, line 5: Running locally `npx prettier --write .` is giving me an extra new line here

nprimo (changes requested):
There is a small format issue with `prettier` where I have added the comment - please, let me know if it's not the same for you

subjects/office_worker/README.md
mikysett: done

nprimo (approved):
LGTM!

PR #1569, opened 2022-11-14
DEV-3809 create new rust exercise - dress_code

nprimo:


subjects/dress_code/README.md
```diff
@@ -0,0 +1,63 @@
+## dress_code
+
+### Instructions
+
+Create a function called `choose_outfit` that receives the following input:
+- a `formality_level` as an `Option<u32>`;
```
mikysett, line 6: - First letter of a line should be Uppercase.
- List items should end with `"."` and not `";"`.

Both of those are just style conventions that I follow, they are not 100% consistent in the piscine, still for new exercises I'm following it quite closely.

Notice this applies to all other list items also.
(special names like `jacket` do not require to have upper case)

subjects/dress_code/README.md
```diff
@@ -0,0 +1,63 @@
...
+fn main() {
+    println!("My outfit will be: {:?}", choose_outfit(Some(0), Ok("Dear friend, ...")));
+}
+```
+
+And its output:
+```console
+$ cargo run
+My outfit will be: Outfit { jacket: Black, hat: Fedora }
+...
```
mikysett, line 57: This line and the next one can be removed

mikysett (changes requested):
couple little details

mikysett (approved):
LGTM!

PR #1581, opened 2022-11-17
DEV-3808 new rust exercise - get_document_id

nprimo:


mikysett (changes requested):
Slightly modify to account the changes in the tests and solution:

- remove the `id` field to reduce useless complexity
- change the u32 returned by errors to better show it is an unrelated id
- use `println` with the `"{}, id"` syntax or `"{:?}, id" for consistency

mikysett (approved):
perfect

PR #1582, opened 2022-11-17
DEV-3810 add new rust exercise - prime_checker

nprimo:


subjects/prime_checker/README.md
```diff
@@ -0,0 +1,43 @@
...
+### Expected Function and structure
+
+```rust
+#[derive(PartialEq, Eq, Debug)]
+pub enum PrimeErr {
+    Even,
+    Divider(u32),
+}
+
+pub fn prime_checker(nb: u32) -> Option<Result<u32, PrimeErr>> {}
```
mikysett, line 20: It would be nice if by giving more details in the previous part we could hide the return type from this prototype, the student will then be required to know and understand the basic behavior of `Option` and `Result` which are the focus of this exercise.

subjects/prime_checker/README.md
```diff
@@ -0,0 +1,43 @@
+## prime_checker
+
+### Instructions
+
+Create a **function** `prime_checker` that takes an `u32` and check if it is a prime number.
+
+The result will be `None` if the argument is less or equal one otherwise, it will return the result with the prime number inside if the number is prime, or it will return an `enum` `PrimeErr` if the `u32` is not prime.
```
mikysett, line 7: Here maybe let's say "the function will return an `Option`. It will be `None` if... and `Result` otherwise".
Then we could specify the `Result` values.

subjects/prime_checker/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+Here is a program to test your function:
+
+```rust
+use prime_checker::*;
+
+fn main() {
+    println!("Is {} prime? {:?}", 2, prime_checker(2));
+    println!("Is {} prime? {:?}", 14, prime_checker(14));
+}
```
mikysett, line 33: maybe we could add one example with a very big number, so students will know we require a little bit of optimization on their code (just couple of smart tricks and not testing every case).

mikysett (changes requested):
I've suggested to hide some code in order to challenge more the students on the topic.

subjects/prime_checker/README.md
nprimo: I will add an extra line to explicitly ask for a certain degree of optimization

subjects/prime_checker/README.md
```diff
@@ -4,9 +4,11 @@
 
 Create a **function** `prime_checker` that takes an `u32` and check if it is a prime number.
 
-The result will be `None` if the argument is less or equal one otherwise, it will return the result with the prime number inside if the number is prime, or it will return an `enum` `PrimeErr` if the `u32` is not prime.
+The result will be `None` if the argument is less or equal one otherwise, it will return a `Result`.
```
mikysett, line 7: the `,` should be after "one" and before "otherwise".
"... ils less or equal one, otherwise it will return ..."

subjects/prime_checker/README.md
```diff
@@ -4,9 +4,11 @@
...
+If the `u32` is prime, the function will return an`Ok(u32)`; for any other case it will return an `enum` `PrimeErr`.
```
mikysett, line 8: - Missing a space for "an`Ok(u32)`"
- To use ";" seems very strange to me, I suggest to use "." or "," as ";" is nowadays almost never used.

subjects/prime_checker/README.md
```diff
@@ -4,9 +4,11 @@
...
+The `enum` `PrimeErr` will be `Even` if the number is a multiple of two or `Divider(u32)` where the `u32` is the smallest divider of the number.
 
-The `enum` `PrimeErr` will be `Even` if the number is even or `Divider(u32)` where the `u32` is the smaller divider of the number if it is not prime.
+> Your solution should be optimized to a certain degree
```
mikysett, line 11: - Missing a "." at the end of the line.

subjects/prime_checker/README.md
```diff
@@ -17,7 +19,7 @@ pub enum PrimeErr {
     Divider(u32),
 }
 
-pub fn prime_checker(nb: u32) -> Option<Result<u32, PrimeErr>> {}
+pub fn prime_checker(nb: u32) -> Option<...> {}
```
mikysett, line 22: What do you think if we keep also the `Option` out of the picture?
Maybe using comments to avoid confusion:
```rust
pub fn prime_checker(nb: u32) -> /* Implement return type */ {}
```

mikysett (changes requested):

PR #1584, opened 2022-11-18
Make instructions more explicit

Pashtetium:
By reading current instructions, student might get wrong impression, that optional starting index is not a thing for `lastIndexOf` function. Which is not true, judging by tests for this exercise


eslopfer:
Hi @Pashtetium the instructions say: 
-`lastIndexOf`: which works just like your `indexOf` function, 

The optional parameter is required for `indexOf`, so it is required as well for `lastIndexOf`,  there is no ambiguity here, and no need to repeat the instructions twice.

I'll close the pull request since it requires no further action, please feel free to open a new one if you find any other issue.

PR #1585, opened 2022-11-18
More clear instructions

Pashtetium:


eslopfer (approved):
Hi @Pashtetium.

You are right. Since the built-in slice doesn't take the array/string as first parameter it might be an unfair ask to guess that it should be passed as first argument. The extra instructions make sense.

Thanks for your feedback.

LGTM :rocket:

PR #1594, opened 2022-11-23
DEV-3806-exam-02-copy-remake-exercise

eslopfer:
I renamed it to make the name be more accurate with the description


subjects/own-and-return/README.md
```diff
@@ -0,0 +1,42 @@
...
+Create a function `own_and_return` which takes ownership of the struct, and returns the string.
+
+### Expected functions
+
+```rust
+pub struct Film {
+    pub name: String,
+}
+
+pub fn only_return(your_parameter) {
```
mikysett, line 16: here usually we put unimplemented code in inline comments, also the return type should be explicit or we should write it must be implemented (here it just seems we do not expect any return).

Actually I think we can say the return is a `String` since there is nothing special about that.

I suggest the follow:
```rust
pub fn only_return(/* to be implemented */) -> String {
```

subjects/own-and-return/README.md
```diff
@@ -0,0 +1,42 @@
...
+}
+
+pub fn own_and_return(your_parameter) {
```
mikysett, line 19: same than `only_return`

subjects/own-and-return/README.md
```diff
@@ -0,0 +1,42 @@
...
+}
+
+
+```
+
+### Usage
+
+Here is a possible program to test your function :
+
+```rust
```
mikysett, line 29: I've initially tough not to give `Usage` for this exercise, still I think you are right and we may give one, but the actual one is not a working program and do not follow our conventions about unimplemented code.

What do you think about this:
```rust
use own_and_return::*;

pub struct Film {
    pub name: String,
}

fn main() {
    let my_film = Film { name: "Terminator" };

    println!("{}", only_return(/* to be implemented */));
    println!("{}", only_return(/* to be implemented */));

    println!("{}", own_and_return(/* to be implemented */));

    // If any of the two following lines is uncommented the program should not compile.
    //  println!("{}",  only_return(\* to be implemented *\));
    //  println!("{}",  own_and_return(\* to be implemented *\));
}
```

Then you can also give an expected output:
And its output:

```console
$ cargo run
Terminator
Terminator
Terminator
$
```

mikysett (changes requested):
For the exercise name, I think it may stay the same even if the function names changes.
For the function names, you may also ignore it and leave them as they are as you see fit.

subjects/own-and-return/README.md
```diff
@@ -0,0 +1,42 @@
+## own_and_return
+
+### Instructions
+
+Create a function `only_return` that takes a `struct` Film that has one Field of type `String` and returns the string without owning it.
```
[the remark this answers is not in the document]
eslopfer: Thanks, Done.

subjects/own-and-return/README.md
eslopfer: Thanks. Done.

subjects/own-and-return/README.md
eslopfer: Thanks. Done

mikysett (changes requested):

subjects/own-and-return/README.md
mikysett: The main function do not correspond to the output, also the commented part is missing (this is an important hint for the student and it gives a way for them to test everything is working as intended)

eslopfer: I added some comments explaining how to test it. It's a bit tricky because of the problem of how to test something that doesn't compile, at  running time :slightly_smiling_face: . I fixed the rest.

subjects/own-and-return/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+pub fn take_film_name(/* to be implemented */) -> String {
+}
+
+
+```
+
+### Usage
+
+Here is a possible program to test your function :
```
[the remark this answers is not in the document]
eslopfer: Good catch, I agree with you, I copy pasted this and it slipped through, I'll be more careful now. It would be nice to fix all the  subjects in the future when priorities allow it :slightly_smiling_face:

subjects/own-and-return/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+```rust
+use own_and_return::*;
+
+
+pub struct Film {
+    pub name: String,
+}
+
+fn main() {
```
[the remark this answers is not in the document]
eslopfer: Done

subjects/own-and-return/README.md
```diff
@@ -0,0 +1,55 @@
+## own_and_return
+
+### Instructions
+
+Create a struct `Film` that has one field `name` of type `String`.
+
+Create the following two functions:
+
+- Create a function `take_film_name` it will return the name and consume the film (you should not be able to reuse it after you passed it to the function).
```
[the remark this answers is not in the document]
eslopfer: Done.

mikysett (approved):
LGTM

subjects/own-and-return/README.md
eslopfer: I'll resolve this conversation since the pull request is approved

PR #1600, opened 2022-11-26
Incorrect table for B donation: B can't donate to O+, O-

Danonika:
Incorrect table for B donation: B can't donate to O+, O-
https://www.inovablood.org/donate-blood/ideal-donation-for-your-blood-type/


Danonika:
@jotapero @davhojt

PR #1602, opened 2022-11-26
Console ouput fix

Danonika:
1)Added 's' for word "parse" in test 3
2)And fixed additional space between words "todo" and "Some"(Right solution prints them together) in test 3


Danonika:
@jotapero

nprimo:
@Danonika thank you for the PR. The current version of the tests print the message as shown in the subject now :)

PR #1607, opened 2022-11-28
DEV-3746 new exercise about simple server handling get request - friend-support

nprimo:


js/tests/friend-support_test.mjs
```diff
@@ -0,0 +1,125 @@
+import { once } from 'node:events'
+import * as cp from 'node:child_process'
```
eslopfer, line 2: Since only spawn is being used, wouldn't it be better to just destructure it `import { spawn } from ...`?

js/tests/friend-support_test.mjs
```diff
@@ -0,0 +1,125 @@
...
+import { mkdir, writeFile, chmod } from 'fs/promises'
+import { tmpdir } from 'node:os'
```
eslopfer, line 4: Imported but not used?

js/tests/friend-support_test.mjs
```diff
@@ -0,0 +1,125 @@
...
+    } catch (err) {
+      body = err
+    }
+    return { status, body, headers, randLastName }
+  }
+
+  return { tmpPath: dir, createFilesIn, sendRequest }
+}
+
+tests.push(async ({ path, ctx }) => {
```
eslopfer, line 42: can ctx be further destructured? `ctx: { server }` at least?

js/tests/friend-support_test.mjs
```diff
@@ -0,0 +1,125 @@
...
+  ctx.server = cp.spawn('node', [`${path}`])
+  const message = await Promise.race([
+    once(ctx.server.stdout, 'data'),
+    Promise.race([
+      once(ctx.server.stderr, 'data').then(String).then(Error),
+      once(ctx.server, 'error'),
+    ]).then(x => Promise.reject(x)),
```
eslopfer, line 49: Can we call x something with meaning? error perhaps?

js/tests/friend-support_test.mjs
```diff
@@ -0,0 +1,125 @@
...
+  ])
+  return message[0].toString().includes(port)
+})
+
+tests.push(async ({ eq, ctx, randStr }) => {
```
eslopfer, line 54: is randStr being used?

eslopfer (changes requested):
It's partially reviewed, I think it would be better to discuss something with you before proceeding further :slightly_smiling_face:

js/tests/friend-support_test.mjs
nprimo: I will remove the unused import statements

js/tests/friend-support_test.mjs
nprimo: I will actually use it to randomize the message to avoid hard-coded solutions. Thank you for the catch :)

js/tests/friend-support_test.mjs
nprimo: The server is now spawned in every test, therefore I kept `ctx` instead of expanding it to the function used from this object

PR #1622, opened 2022-12-05
DEV-4016 - improve manipulate-entries subject

nprimo:
Add the `usage` section in README


mikysett (approved):
The only thing that I see is to add a space at `line 14` between `50` and `grams`.

Otherwise everything is fine and can be merged.

PR #1626, opened 2022-12-06
DEV-3902-head-and-tail

zamazzal:


sh/tests/head-and-tail_test.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+# Unofficial Bash Strict Mode
+set -euo pipefail
+IFS='
+'
+
+echo insecure >> ~/.curlrc
+caddy start &>/dev/null
+
```
nprimo, line 10: I would check if the file has been submitted correctly as it has been done in the exercise `to-git-or-not-to-git` (https://github.com/01-edu/public/blob/master/sh/tests/to-git-or-not-to-git_test.sh)

sh/tests/head-and-tail_test.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
```
nprimo, line 2: should we add a check to ensure the student solution is not a simple `echo`?

Maybe something like
```bash
if [[ $(cat <path-to-student-solution> | grep echo | wc -l) -ne 0 ]];
then
  echo "Bad solution";
fi;
```
Or, viceversa, check that the solution is using `curl`, `head` and `tail` keywords
Or both checks :)

subjects/devops/head-and-tail/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+```console
+$tail file
+<last 10 lines>
+$tail -1 file
+<last 1 line>
+$
+```
+
+May you will need to use `pipe (|)` and `&&`.
```
nprimo, line 50: **SUGGESTION** I will rephrase this as follows:
You will need to use pipes (`|`) and the logical operator `&&`

nprimo (changes requested):

sh/tests/head-and-tail_test.sh
zamazzal: fixed

sh/tests/head-and-tail_test.sh
zamazzal: fixed

subjects/devops/head-and-tail/README.md
zamazzal: they are different solutions to solve this exercise, and it's just a hint

nprimo (commented):
Isn't the `README.md` supposed to be in the main repo `public/subjects`?

PR #1629, opened 2022-12-06
ascii-art-output: fix shell prompt at the end of the examples

robcat:


nprimo:
@robcat, thank you for your input. This issue should be soon fixed by [this PR](https://github.com/01-edu/public/pull/2472)

Issue #1618, opened 2022-12-03
math-skills subject - instructions


fulverin:
math-skills
---
Hi,
"This data represents a graph in which the values of the x axis are the number of the lines (0, 1, 2, 3, 4, 5 ...) and the values of the y axis are the actual numbers (189, 113, 121, 114, 145, 110...)."
Sorry but that does not make sense: a graph represent data , not the other way or at least not in this case; and we do not even need a graph at all...
I would rather **suggest** using the line numbers (0, 1, 2, 3, 4, 5 ...) as indices of the following **actual data values** (189, 113, 121, 114, 145, 110...)
but order is not relevant here, since calculation does not care about it.
I m not sure if it is supposed to help people with no math background, but that confuses the ones that have one for sure.
Regards

eslopfer:
Hi @fulverin and @robcat, thanks for your inputs.

This issue is being addressed and will be solved soon.

PR #1634, opened 2022-12-07
DEV-4023 rephrase presentation of the input file

eslopfer:
As suggested by user @robcat: 

Spurious references about graphs, x-values, and y-values are removed because they are not connected at all with the subject. The term "statistic population" is introduced, from which it can be inferred that the program should calculate the population variance (and not the sample variance) should fix issue #1618


davhojt (approved):
LGTM.

PR #1635, opened 2022-12-07
Proposing change to lem-in audit question

October45:
# Question edit
Changes the lem-in audit questions concerning the "good" examples to be more free in interpretation, so that you don't need the exact same number of turns, but it can be less.
# Question addition
Added the number of turns needed according to the example in plain text after the question so that there's no need to count.


mikysett:
Another PR created with few more little changes: [#1639](https://github.com/01-edu/public/pull/1639)

PR #1648, opened 2022-12-13
DEV-3896-introduction

zamazzal:
…o teach student how to push in gitea repo


MSilva95 (changes requested):
It looks ok, but the name we should change to "hello-devops" to keep the same consistency as the others.

MSilva95 (approved):

PR #1651, opened 2022-12-13
 DEV-3909 add subject, test and solution for the exercise easy-perm

MSilva95:


sh/tests/easy-perm_test.sh
```diff
@@ -0,0 +1,17 @@
...
+# Unofficial Bash Strict Mode
+set -euo pipefail
+IFS='
+'
+script_dirS=$(cd -P "$(dirname "$BASH_SOURCE")" &>/dev/null && pwd)
+
+chmod 303 easy-perm/*
+
+challenge() {
+    submitted=$(bash "$script_dirS"/student/easy-perm.sh && ls -l "$1")
```
zamazzal, line 12: you have to check the permission only
with ls -l you will compare group of files infos, which will impact the tests results.

zamazzal (changes requested):

sh/tests/easy-perm_test.sh
MSilva95: the exercise also demands that the file struct is the exact same one as ours so i used ls -l because of that should we change it?

sh/tests/easy-perm_test.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+# Unofficial Bash Strict Mode
+set -euo pipefail
+IFS='
+'
+script_dirS=$(cd -P "$(dirname "$BASH_SOURCE")" &>/dev/null && pwd)
+
+challenge() {
+    submitted=$(ls -l "$1" && bash "$script_dirS"/student/easy-perm.sh)
```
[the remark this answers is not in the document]
MSilva95: done

zamazzal (approved):

sh/tests/easy-perm_test.sh
zamazzal: fixed by MS
MS in discord: right, i will use awk to filter in the test, Thanks 👍

PR #1655, opened 2022-12-15
DEV-3918-auto-exec-bin

zamazzal:
…bin exercise

### DIFF
--- a/sh/tests/auto-exec-bin_test.sh
+++ b/sh/tests/auto-exec-bin_test.sh
@@ -14,23 +14,20 @@ setupbin() {
         echo "bin already exists!"
     else
         mkdir -p ~/myBins
-        curl -o ~/myBins/01exec https://assets.01-edu.org/devops-branch/01exec
-        chmod +x ~/myBins/01exec
-        echo "bin installed!"
+        echo "echo Hello 01 Scripting Pool" > $HOME/myBins/01exec
     fi
+    chmod +x $HOME/myBins/01exec
 }
 
 challenge() {
+    OLD_PATH=$PATH
     # run soultion script
-    export PATH=""
-
-   	$(bash "$script_dirS"/$FILENAME)
+    source "$script_dirS"/$FILENAME
 	submitted=$(cd / && 01exec)
 
+    PATH=$OLD_PATH
     # run student script
-    export PATH=""
-
-	$(bash "$script_dirS"/solutions/auto-exec-bin.sh)
+	source "$script_dirS"/solutions/auto-exec-bin.sh
 	expected=$(cd / && 01exec)
 
     # diff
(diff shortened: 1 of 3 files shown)


sh/tests/auto-exec-bin_test.sh
```diff
@@ -0,0 +1,56 @@
...
+
+# Unofficial Bash Strict Mode
+set -euo pipefail
+IFS='
+'
+
+FILENAME="student/auto-exec-bin.sh"
+script_dirS=$(cd -P "$(dirname "$BASH_SOURCE")" &>/dev/null && pwd)
+
+setupbin() {
```
MSilva95, line 11: this is checking if auto-exec-bin.sh exists
If it exist the myBins folder will not be created right?

MSilva95 (changes requested):
I get this as soon as i run the test:

```
./auto-exec-bin_test.sh 
bin already exists!
./auto-exec-bin_test.sh: line 26: bash: No such file or directory
```
and i have all the folders and solution in the correct place. im not sure what is happening...

sh/tests/auto-exec-bin_test.sh
zamazzal: yes

zamazzal: yes i have a bug here, I have a uncorrected condition
fixed by checking for the binary

sh/tests/auto-exec-bin_test.sh
```diff
@@ -0,0 +1,57 @@
...
+        mkdir -p ~/myBins
+        curl -o ~/myBins/01exec https://assets.01-edu.org/devops-branch/01exec
+        chmod +x ~/myBins/01exec
+        echo "bin installed!"
+    fi
+}
+
+challenge() {
+    # run soultion script
+    export PATH=""
```
MSilva95, line 25: when i run the test i get this error:
"./auto-exec-bin_test.sh: line 28: 01exec: command not found"
The new path is never added as you can see in the print, that is my current path printed after the student script run, and it doesn't add the "~/myBins" to the path. so the test is stuck there. 
image

MSilva95: can you run the test in your computer without this error?

MSilva95 (changes requested):
test is stuck because the new path is not added

sh/tests/auto-exec-bin_test.sh
nprimo: I replaced the 01exec with a simple bash script - the one used before was not running on my machine (Ubuntu 22.04)

MSilva95: Nice!

MSilva95 (approved):
LGTM

PR #1660, opened 2022-12-18
DEV-3919-my-ls

zamazzal:


sh/tests/solutions/my-ls.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+alias my-ls="ls -lgSsa"
```
MSilva95, line 3: with this solution the alias is never created in the student machine

sh/tests/my-ls_test.sh
```diff
@@ -0,0 +1,43 @@
...
+# Unofficial Bash Strict Mode
+set -euo pipefail
+IFS='
+'
+
+FILENAME="student/my-ls.sh"
+script_dirS=$(cd -P "$(dirname "$BASH_SOURCE")" &>/dev/null && pwd)
+
+challenge() {
+    unalias my-ls
```
MSilva95, line 12: the test gets stuck here even using your solution.
"./my-ls_test.sh: line 12: unalias: my-ls: not found
"
I think is because the alias is never saved in the bashrc file with that solution

MSilva95 (changes requested):
the test is getting stuck with the given solution

subjects/devops/my-ls/README.md
```diff
@@ -0,0 +1,61 @@
+## my-ls
```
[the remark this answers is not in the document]
zamazzal: changed to custom-ls

subjects/devops/my-ls/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Instructions
+
+Create the script `my-ls.sh` wich will create an alias `my-ls`.
```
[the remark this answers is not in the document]
zamazzal: fixed

subjects/devops/my-ls/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+The alias `my-ls`:
```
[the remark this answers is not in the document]
zamazzal: done

sh/tests/my-ls_test.sh
zamazzal: yes fixed cause a new env will created to exec this script, so the alias is not exists

sh/tests/solutions/my-ls.sh
zamazzal: not required, the alias will be created in the working env, and its enough for the testing
but i add a part in the hints section to teach the student how to create a permanently alias

sh/tests/my-ls_test.sh
MSilva95: still the test doesn't run:
image
and i'm using your solution as the student solution so they are the same

MSilva95 (changes requested):
The test doesn't run properly, it says that the alias doesn't exist.

mikysett:
By default aliases are not expanded in script (they only are in interactive mode).
To modify that behavior I've added the proper shell option at the start `shopt -s expand_aliases`.
Also the scripts had to be ran with `source` and not with `bash` otherwise the aliases would be taken by the inner bash and not by the caller.

PR #1666, opened 2022-12-20
DEV-4069 improve ascii-art-color audit

nprimo:
Replace double quotes with single quotes to avoid special char effect


davhojt (approved):
LGTM.

PR #1671, opened 2022-12-20
DEV-3911-file-details add subject, test and solution for the exercise

MSilva95:


sh/tests/file-details_test.sh
```diff
@@ -0,0 +1,17 @@
...
+script_dirS=$(cd -P "$(dirname "$BASH_SOURCE")" &>/dev/null && pwd)
+
+challenge() {
+    submitted=$(cd "$1" && bash "$script_dirS"/student/file-details.sh)
+    expected=$(cd "$1" && bash "$script_dirS"/solutions/file-details.sh)
+
+    diff <(echo "$submitted") <(echo "$expected")
+}
+
+challenge hard-perm
```
zamazzal, line 17: hard-perm folder is not included in the pr, I didn't found it in the repo

zamazzal (changes requested):

sh/tests/file-details_test.sh
MSilva95: Yes but it will be because it was added in the hard perm exercise

zamazzal (changes requested):
the goal of this exercise is not to show the files with specific format
the goal of this exercise to create files and folders with specific details
so the student can learn how to change files details, like last modification time, file size, ...
check this: https://01talent.atlassian.net/browse/DEV-3911
but cause you already make a great effort to create this exercise, you can keep this exercise and create another exercise to achieve this goal.

zamazzal (approved):

PR #1683, opened 2022-12-27
feat(right): add subject, test and solution for the exercise

MSilva95:


sh/tests/solutions/right.sh
```diff
@@ -0,0 +1 @@
+ls | grep -v "\.txt" >filtered_files_sol.txt
```
mikysett, line 1: here the file should become `filtered_files.txt` after the refactor

subjects/devops/right/README.md
```diff
@@ -0,0 +1,28 @@
...
+$ ./right.sh
+$ cat filtered_files.txt
+sample2
+sample4
+$
+```
+
+### Hints
+
+command1 | command2 > output_file
```
mikysett, line 23: May could be made cleared if wrapped `command1 | command2 > output_file`
or even a code bloc
```console
command1 | command2 > output_file
```

subjects/devops/right/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+Here, command1 is the command that generates the output you want to parse, and command2 is the command that parses the output. The output of command2 will be redirected to the file output_file using the > operator.
```
mikysett, line 25: command1 -> `command1`
command2 -> `command2` x2
output_file -> `output_file`
also the operator > should be `>`

sh/tests/right_test.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+# Unofficial Bash Strict Mode
+set -euo pipefail
+IFS='
+'
+
+FILENAME="student/right.sh"
+script_dirS=$(cd -P "$(dirname "$BASH_SOURCE")" &>/dev/null && pwd)
+
```
mikysett, line 10: Here we could add the necessary boilerplate for the script to make tests, like:
```sh
mkdir right
touch right/a_file.txt right/txt_file right/txt right/example right/example.doc right/example.TXT right/example.txtnot
```

like so we don't have example files for tests in the repo, and we see the logic more easily in the testing script.

mikysett (changes requested):
The subject is perfect and just has some styles improvements.

Still I think there is room for improvement in the following fields:
- Structure the test so the solution doesn't need to output on a different file than the student
```sh
    $(cd "$1" && bash "$script_dirS"/$FILENAME)
    submitted=$(cat $1/filtered_files.txt)
    rm $1"/filtered_files.txt"
    $(cd "$1" && bash "$script_dirS"/solutions/right.sh)
    expected=$(cat $1/filtered_files.txt)
```
- Removing the `right` directory from the repo and create it on the go for the exercise needs (for a cleaner repo and a more straight forward dependency management (see inline comment in the test)

subjects/devops/right/README.md
MSilva95: done

subjects/devops/right/README.md
MSilva95: done

sh/tests/right_test.sh
MSilva95: fixed

PR #1686, opened 2022-12-28
DEV-3925-left: add subject, test and solution for the exercise

MSilva95:


subjects/devops/left/README.md
```diff
@@ -0,0 +1,29 @@
...
+```console
+$./left.sh
+21
+```
+
+### Hints
+
+To pass the content of a file to a command in a shell, you can use the "<" operator followed by the name of the file. For example, to pass the content of a file named "input.txt" to the cat command, you can use the following command:
+
+`cat < input.txt`
```
nprimo, line 18: I feel the example command should be something like `< input.txt cat` instead of the one presented. The result is the same but it shows better `<` metacharacter's effect

sh/tests/solutions/left.sh
```diff
@@ -0,0 +1 @@
+wc -l <facts
```
nprimo, line 1: As a general information, the student solution can pass without using the redirection

sh/tests/left_test.sh
```diff
@@ -0,0 +1,30 @@
...
+script_dirS=$(cd -P "$(dirname "$BASH_SOURCE")" &>/dev/null && pwd)
+
+FILENAME="../student/left.sh"
+cd left
+
+# True if FILE exists and is a regular file
+if [ -f ${FILENAME} ]; then
+    # FILE exists and it's not empty
+    if [ -s ${FILENAME} ]; then
+        if [[ $(cat $FILENAME | grep echo | wc -l) -ne 0 ]]; then
```
nprimo, line 16: Should we add a check for the symbol `<` - without it the solution cannot pass?

sh/tests/left_test.sh
```diff
@@ -0,0 +1,30 @@
...
+            echo "echo is not allowed in this exercise!"
+            exit 1
+            submitted=$(bash $FILENAME)
+            expected=$(bash ../solutions/left.sh)
+            diff <(echo "$submitted") <(echo "$expected")
+        fi
```
nprimo, line 22: this line should be moved to right after `exit 1`

sh/tests/left_test.sh
```diff
@@ -0,0 +1,30 @@
...
+    else
+        echo "The file exist but is empty"
+        exit 1
+    fi
+else
+    echo "File does not exist"
+    exit 1
+fi
```
nprimo, line 30: I suggest the following refactoring: 

```bash
#!/usr/bin/env bash

# Unofficial Bash Strict Mode
set -euo pipefail
IFS='
'
script_dirS=$(cd -P "$(dirname "$BASH_SOURCE")" &>/dev/null && pwd)

FILENAME="../student/left.sh"
cd left

if [[ $(cat $FILENAME | grep echo | wc -l) -ne 0 ]]; then
    echo "echo is not allowed in this exercise!"
    exit 1
fi

# optional
if [[ $(cat $FILENAME | grep '<' | wc -l) -lt 1 ]]; then
    echo "your solution must use the < symbol!"
    exit 1
fi

submitted=$(bash $FILENAME)
expected=$(bash ../solutions/left.sh)


```

nprimo (changes requested):
I left some comments about the current test

sh/tests/solutions/left.sh
MSilva95: i added a restriction

sh/tests/left_test.sh
MSilva95: added

sh/tests/left_test.sh
MSilva95: i changed the test to your suggestion it's working correctly now

nprimo (approved):
LGTM!

PR #1688, opened 2022-12-29
DEV-4102 move guests folder to public repository

eslopfer:
- Moved folder to public repo.
- Updated links to new location.


davhojt (changes requested):
It's better not to have resources in the subjects directory. They can be:
- inside an exercise. i.e. `/subjects/gatecrashers/resources/guests.zip`, and different exercises can reference it.
- inside each exercise (essentially duplicated to many places).

I'm actually not sure which of those is best.

eslopfer:
I put the folder in the first exercise that it is required, and a link to it on the next exercises, since I think is better to avoid repetition.

A potential trade-off could be that being only on one place, if that exercise got deleted in the future, care should be taken to locate the resource somewhere else, and the links in the rest of the exercises should be updated accordingly.

Let me know if you prefer a different solution.

davhojt (approved):

PR #1695, opened 2022-12-30
DEV-3932-lamp-in-the-dark: add subject, test and solution for the exercise

MSilva95:


nprimo:
No added value - very similar to `burial` exercise

PR #1697, opened 2023-01-03
Dev 3946 array selector

nprimo:


subjects/devops/array-selector/README.md
```diff
@@ -0,0 +1,95 @@
+## array-selector
+
+### Instructions
+
+Create a script `array-selector.sh`, which will have an array declared with the
+following values (in this order): `red`, `blue`, `green`, `white`, `black`.
+
+When executed, the script will try to print the element in the position specified in the first argument passed to the script.
+
+The script should return `Error` when:
```
MSilva95, line 10: Just an opinion:
```
The script should return `Error` when:

- the number of given arguments is different from one,

- the given argument is not a number,

- the given argument is a number outside the range of the array.
```

subjects/devops/array-selector/README.md
```diff
@@ -0,0 +1,95 @@
+## array-selector
+
+### Instructions
+
+Create a script `array-selector.sh`, which will have an array declared with the
+following values (in this order): `red`, `blue`, `green`, `white`, `black`.
+
+When executed, the script will try to print the element in the position specified in the first argument passed to the script.
```
MSilva95, line 8: the element at the position

sh/tests/array-selector_test.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
```
MSilva95, line 1: this file is not executable.
Also the test passes with an empty solution

sh/tests/solutions/array-selector.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
MSilva95, line 1: this file is not executable

sh/tests/array-selector_test.sh
```diff
@@ -0,0 +1,33 @@
...
+
+challenge() {
+    submitted=$(bash "$script_dirS"/student/array-selector.sh $1)
+    expected=$(bash "$script_dirS"/solutions/array-selector.sh $1)
+
+    diff <(echo "$submitted") <(echo "$expected")
+}
+
+# Test with numbers - out of range included
+for num in {0..6}
```
MSilva95, line 19: the for is not working i thin it is because it starts at 0 and 0 is supposed to give you error and it is not working 
If you do the for like so:
```
for num in {1..6}; do
    challenge $num
done
```
it will be a valid test for the correct values of the array, but it will still fail for the error part

subjects/devops/array-selector/README.md
```diff
@@ -0,0 +1,95 @@
+## array-selector
+
+### Instructions
+
+Create a script `array-selector.sh`, which will have an array declared with the
```
MSilva95, line 5: Explain somewhere that this array will work different than normal as it will start at position 1 instead of 0.

subjects/devops/array-selector/README.md
nprimo: I like it better, I will change it like you suggested :)

sh/tests/solutions/array-selector.sh
nprimo: I will change this just to avoid weird behaviour, but it should not be a problem since the test script is running `bash <filename>`

subjects/devops/array-selector/README.md
nprimo: I have added a sentence to make it more explicit - this can be extrapolated as well from the given examples

sh/tests/array-selector_test.sh
nprimo: The empty solution issue was related to the `exit` in the previous solution version - now it should be fixed

MSilva95 (changes requested):

MSilva95 (approved):
LGTM!

PR #1698, opened 2023-01-04
DEV-3943-plus

eslopfer:
- Solution, tests and description for exercise.


subjects/devops/plus/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+And `$(...)` like this:
+
+```console
+$ echo "This is the content of a file: $(cat file.txt)"
+This is the content of a file: <content of the file>`
+$
+```
+
+### Notions
```
MSilva95, line 39: We are using this `### References` instead

subjects/devops/plus/README.md
```diff
@@ -0,0 +1,43 @@
+## plus
+
+### Instructions
+
+In this exercise, you will make a script that will take two arguments from the command line, add them and output the result.
```
MSilva95, line 5: Give the exact name of the script here 
"Create a script `plus.sh` "

subjects/devops/plus/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Usage
+
+```console
```
MSilva95, line 9: We are not using the cat -e flag like in the go and etc... so if you decide to use it here add a hint to the hint section explaining the flag as its the first time it will appear.

MSilva95 (changes requested):
Both of the scrips are not executable, you need to make it executable in order to run the test

subjects/devops/plus/README.md
eslopfer: Thanks. Changed.

subjects/devops/plus/README.md
eslopfer: Thanks. Done.

subjects/devops/plus/README.md
eslopfer: That's right. I'll remove it since it's not needed. Thank you.

PR #1701, opened 2023-01-05
DEV-3944 division

eslopfer:


sh/tests/division_test.sh
```diff
@@ -0,0 +1,24 @@
...
+    expected=$(bash "$script_dirS"/solutions/division.sh $1 $2)
+
+    diff <(echo "$submitted") <(echo "$expected")
+}
+
+challenge "10" "2"
+challenge "4" "2"
+challenge "0.5" "0.5"
+challenge "5" "2"
+challenge "0.5"
```
MSilva95, line 21: This is not testing properly it gives this error
`./division_test.sh: line 11: $2: unbound variable`

sh/tests/division_test.sh
```diff
@@ -0,0 +1,24 @@
...
+challenge "foo" "bar"
```
MSilva95, line 22: this never fails the test as well even if the student don't have the correct error message
so i guess its not even testing

sh/tests/division_test.sh
```diff
@@ -0,0 +1,24 @@
...
+challenge "1" "0"
```
MSilva95, line 23: here as well

sh/tests/division_test.sh
```diff
@@ -0,0 +1,24 @@
...
+challenge "10000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000" "2"
```
MSilva95, line 24: and this gives an odd number
`50000000000000000000000000000000000000000000000000000000000000000000\ 00000000000000000000000000000000`
is it normal?

sh/tests/solutions/division.sh
```diff
@@ -0,0 +1,28 @@
...
+if [ $# -ne 2 ]
+then
+    echo "Error: two numbers must be provided"
+    exit 1
+fi
+
+# Check if the arguments are numeric
+if ! [[ $1 =~ ^-?[0-9]*\.?[0-9]+$ ]] || ! [[ $2 =~ ^-?[0-9]*\.?[0-9]+$ ]]
+then
+    echo "Error: both arguments must be numeric"
```
MSilva95, line 13: test not testing this

sh/tests/solutions/division.sh
```diff
@@ -0,0 +1,28 @@
...
+    exit 1
+fi
+
+# Check if the second argument is not 0
+if [ $(echo "$2 == 0" | bc) -eq 1 ]
+then
+    echo "Error: division by zero is not allowed"
```
MSilva95, line 20: test not testing this

sh/tests/solutions/division.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+
+# Check if two arguments were provided
+if [ $# -ne 2 ]
+then
+    echo "Error: two numbers must be provided"
```
MSilva95, line 6: test not testing this

MSilva95 (changes requested):

sh/tests/division_test.sh
eslopfer: Thanks. I added a condition to test with one argument when one is provided and with two when there is two.

sh/tests/division_test.sh
eslopfer: Thanks, I realised that I was testing in a wrong way, because the solution was exiting with 1 it wasn't outputting the diff and it would always fail. I was testing it, but I suspect that because of the exit status 1 it was never getting to the point to display the diff. I'll modify the exercise to make it work with an exit 0

sh/tests/division_test.sh
eslopfer: I think it's normal, I presume it's the default fold from the terminal. The result is the correct number if you remove the `\`

sh/tests/division_test.sh
eslopfer: Same as the comment above

sh/tests/solutions/division.sh
eslopfer: It is, it is explained above.

sh/tests/solutions/division.sh
eslopfer: It is, it is explained above.

sh/tests/solutions/division.sh
eslopfer: It is, it is explained above.

sh/tests/division_test.sh
MSilva95: cool, i just found it odd so i asked.

MSilva95 (approved):
LGTM

PR #1704, opened 2023-01-06
DEV-3948-comparator

MSilva95:


sh/tests/solutions/comparator.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
```
nprimo, line 2: Should the script test if it is run with the correct number of arguments?

subjects/devops/comparator/README.md
```diff
@@ -0,0 +1,21 @@
+## comparator
+
+### Instructions
+
+Create a script `comparator.sh` which will verify if the first given argument is bigger, smaller or equal than the second given argument.
+You must print te output like the example in usage.
+
```
nprimo, line 7: If no checks about the argument passed are required, shouldn't this be specified in the subject? Something like
> the script will assume that it will be always run with two numbers as arguments

nprimo (changes requested):
I left a comment to add clarity for the cases where the script is run with not valid arguments - I would go by specifying in the subject that the student can assume that the script will always be run with 2 numbers as arguments

sh/tests/solutions/comparator.sh
MSilva95: Done

subjects/devops/comparator/README.md
MSilva95: Done

subjects/devops/comparator/README.md
```diff
@@ -0,0 +1,21 @@
+## comparator
+
+### Instructions
+
+Create a script `comparator.sh` which will accept only two numbers as arguments and it will verify if the first given argument is bigger, smaller or equal than the second given argument.
+You must print te output like the example in usage.
```
nprimo, line 6: there is a small typo - `te` should be `the`

sh/tests/solutions/comparator.sh
```diff
@@ -0,0 +1,13 @@
...
+if [ "$#" -ne 2 ]; then
+    echo "Error: script requires two arguments to run"
```
nprimo, line 4: This error message is not specified in the subject

sh/tests/solutions/comparator.sh
```diff
@@ -0,0 +1,13 @@
...
+elif ! [[ $1 =~ ^-?[0-9]*\.?[0-9]+$ ]] || ! [[ $2 =~ ^-?[0-9]*\.?[0-9]+$ ]]; then
+    echo "Error: Only two numeric arguments are acceptable"
```
nprimo, line 6: This error message is not specified in the subject

subjects/devops/comparator/README.md
nprimo: The subject should include some information on what should happen when there is an error

sh/tests/comparator_test.sh
```diff
@@ -0,0 +1,34 @@
...
+script_dirS=$(cd -P "$(dirname "$BASH_SOURCE")" &>/dev/null && pwd)
+
+challenge() {
+    if [ $# -ne 2 ]; then
+        submitted=$(bash "$script_dirS"/student/comparator.sh $1)
+        expected=$(bash "$script_dirS"/solutions/comparator.sh $1)
+    else
+        submitted=$(bash "$script_dirS"/student/comparator.sh $1 $2)
+        expected=$(bash "$script_dirS"/solutions/comparator.sh $1 $2)
+        diff <(echo "$submitted") <(echo "$expected")
```
nprimo, line 19: the `diff` should be after the `if`- the test is not failing when the number of argument is different then 2 and there are differences

nprimo (changes requested):
There is a small typo + there is a bug in the test script

to simplify, the `challenge` function could be refactored as below
```bash
challenge () {
challenge() {
    submitted=$(bash $script_dirS/student/comparator.sh "$@")
    expected=$(bash $script_dirS/solutions/comparator.sh "$@")
}
```

nprimo (approved):
LGTM!

PR #1706, opened 2023-01-09
DEV-3951-calculator

mikysett:
Add a new exercise to scripting piscine

Notes:
- For the tests the unofficial strict mode have been turned off because we need to test failing cases with non-zero statuses returned.
- `challenge` function have been modified to check also status codes returned by the script
- Also `challenge` now accepts a variadic number of arguments
- `diff` has now more context, giving to the student the arguments passed to the failing command


subjects/devops/calculator/README.md
```diff
@@ -0,0 +1,62 @@
...
+$ ./calculator.sh -1 - 10
+-11
+$
+```
+
+### Error handling
+
+All errors will print a specific message on **stderr** (ending with a newline) and returns a specific non-zero value:
+- Wrong number of arguments: `"Error: expect 3 arguments"`, returns `1`.
+- Division by 0: `"Error: division by 0"`, exit with `2`.
```
MSilva95, line 38: This to exits are changed in the solution.
is it be like it is said in the readme or like it is in the solution?

image

subjects/devops/calculator/README.md
```diff
@@ -0,0 +1,62 @@
...
+print_full_name () {
+    name=$1
+    surname=$2
+    echo $name $surname
+}
+
+print_full_name "Gene" "Mallamar"
+```
+
+> [Bash functions](https://linuxize.com/post/bash-functions/)
```
MSilva95, line 58: If you want to add links you  need to add the section
```
### References
```

subjects/devops/calculator/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+All errors will print a specific message on **stderr** (ending with a newline) and returns a specific non-zero value:
+- Wrong number of arguments: `"Error: expect 3 arguments"`, returns `1`.
+- Division by 0: `"Error: division by 0"`, exit with `2`.
+- Invalid operator: `"Error: invalid operator"`, exit with `3`.
+- Invalid number(s): `"Error: invalid number"`, exit with `4`.
+
+> Negative numbers are also a valid input.
+
+### Hints
```
MSilva95, line 44: Here in the hints, normally we are giving more info in the commands they will probably use, like an intro to what is the command an an example of usage.
In this piscine we are assuming that the students will not have programming background so we need to help as much as possible without giving the answer of course.
So in this case you can add an example of some commands usages that they need to know.

sh/tests/solutions/calculator.sh
```diff
@@ -0,0 +1,66 @@
...
+}
+
+
+if [ $# != 3 ]
+then
+    >&2 echo "Error: expect 3 arguments"
+    exit 1
+elif ! [[ $1 =~ $number && $3 =~ $number ]]
+then
+    >&2 echo "Error: not a valid number"
```
MSilva95, line 33: different message in the readme
`"Error: invalid number"`

MSilva95 (changes requested):

subjects/devops/calculator/README.md
MSilva95, line 44: show an example of the "case" condition here, as this exercise is supposed to introduce case to the students.
also if you can maybe we should add a verification in the test to check if they use case or not.
basically make case a mandatory condition for this exercise, what do you think?

subjects/devops/calculator/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+Each operator should have its own function named as follow:
+- `+`: `do_add()`.
+- `-`: `do_sub()`.
+- `*`: `do_mult()`.
+- `/`: `do_divide()`.
+
+Each function will receive two arguments, the left number and the right number, and return the result of the operation.
+
+The functions assume the input is valid, so the input should be checked before to call the functions.
```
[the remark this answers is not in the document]
mikysett: done

sh/tests/solutions/calculator.sh
mikysett: updated the solution to match the subject

subjects/devops/calculator/README.md
mikysett: done

subjects/devops/calculator/README.md
mikysett: I've added hints for the `getent` command, which is normally the only new command they will need to use.

mikysett: - I've explicitly added the usage of `case` as a requirement.
- Also now there is an example of `case` usage in the `Hints` paragraph.

The usage of case is now strongly encouraged and it looks mandatory to the student, but in terms of automated tests I don't think it is possible to check that the student actually uses `case`.
If you have an idea on how to add this automated verification in tests I'll be glad to implement it, that could be very interesting to do actually.

MSilva95: Im not sure as well if it is possible, maybe with a grep like we use to check if they dont use echo
```
FILENAME="student/calculator.sh"

if [[ $(cat $FILENAME | grep case | wc -l) -ne 1 ]]; then
    echo "case is mandatory!"
fi
```
but i guess if it is not we can just encouraged on the readme

MSilva95: Yes we should only add info if it is something new like the `getent`, but maybe you forgot to push this change because i don't see any info on `getent` in the readme.

MSilva95 (changes requested):
If it is not possible to check if case is used we can merge like this.

subjects/devops/calculator/README.md
MSilva95: i guess the `getent` is for the other exercise

mikysett: Thanks for the example, I've modified it slightly and added it so we have a test (even if a bit partial) for usage of case statement.

sh/tests/calculator_test.sh
```diff
@@ -0,0 +1,86 @@
...
+
+    diff -U 1 <(echo "$submitted") <(echo "$expected")
+    if [ $? != 0 ]
+    then
+        exit 1
+    fi
+}
+
+# Check if student uses case statement
+if [[ $(cat "$script_dirS"/solutions/calculator.sh | grep case | wc -l) -eq 0 ]]
```
MSilva95, line 29: here you are checking in the solution but it must be the student.
"cat "$script_dirS"/student/calculator.sh "

MSilva95 (changes requested):
The 
- Dockerfile
- enterypoint.sh
- test_in_containor 
must be removed from the branch before merging

sh/tests/calculator_test.sh
mikysett: done

MSilva95 (approved):
LGTM!

PR #1707, opened 2023-01-09
DEV-3954-check-user

mikysett:
add a new exercise to scripting piscine.

Here also the unofficial script mode is turned off in tests because some case scenarios do expect non-zero return statuses

### DIFF
--- a/sh/tests/check-user_test.sh
+++ b/sh/tests/check-user_test.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+
+# set -euo pipefail
+IFS='
+'
+
+script_dirS=$(cd -P "$(dirname "$BASH_SOURCE")" &>/dev/null && pwd)
+
+challenge() {
+    submitted="./check-user.sh $@
+    "
+    expected="./check-user.sh $@
+    "
+    submitted+=$(2>&1 bash "$script_dirS"/student/check-user.sh "$@")
+    expected+=$(2>&1 bash "$script_dirS"/solutions/check-user.sh "$@")
+
+    diff -U 1 <(echo "$submitted") <(echo "$expected")
+}
+
+challenge "-i" "root"
+challenge "-e" "root"
+
+challenge "-i" "unknown_not_found"
+challenge "-e" "unknown_not_found"
+
+challenge
+challenge "-i" "root" "too" "many" "args"
+challenge "-t" "root"
+
--- a/sh/tests/solutions/check-user.sh
+++ b/sh/tests/solutions/check-user.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+IFS='
+'
+
+if [ $# != 2 ]
+then
+    >&2 echo "Error: expect 2 arguments"
+    exit 1
+elif [[ $1 == "-e" ]]
+then
+    user_info=$(getent passwd $2)
+    if [[ $user_info == "" ]]
+    then
+        echo no
+    else
+        echo yes
+    fi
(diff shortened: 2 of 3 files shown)


subjects/devops/check-user/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+> Your results may appear slightly different.
+
+### Error handling
+
+All errors will print a specific message on **stderr** (ending with a newline) and returns a specific non-zero value:
+- Wrong number of arguments: `"Error: expect 2 arguments"`, returns `1`.
+- First argument different from `-e` or `-i`: `"Error: unknown flag"`, exit with `1`.
+
+### Hints
```
MSilva95, line 36: You need to give examples of the getent usage and explain a little what is that command

subjects/devops/check-user/README.md
```diff
@@ -0,0 +1,39 @@
...
+$ ./check-user.sh -i unknown
+$
+```
+
+> Your results may appear slightly different.
+
+### Error handling
+
+All errors will print a specific message on **stderr** (ending with a newline) and returns a specific non-zero value:
+- Wrong number of arguments: `"Error: expect 2 arguments"`, returns `1`.
```
MSilva95, line 33: Why is one written  "returns `1`" and the other is " exit with `1`."?
Can they be written the same way in order to avoid confusion?

sh/tests/check-user_test.sh
```diff
@@ -0,0 +1,29 @@
...
+'
+
+script_dirS=$(cd -P "$(dirname "$BASH_SOURCE")" &>/dev/null && pwd)
+
+challenge() {
+    submitted="./check-user.sh $@
+    "
+    expected="./check-user.sh $@
+    "
+    submitted+=$(2>&1 bash "$script_dirS"/student/check-user.sh "$@")
```
MSilva95, line 14: not sure why but as soon as i format this changes to 
```
submitted+=$(bash 2>&1 "$script_dirS"/student/check-user.sh "$@")
expected+=$(bash 2>&1 "$script_dirS"/solutions/check-user.sh "$@")
```
But it works the same way so no worries here

sh/tests/solutions/check-user.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
```
MSilva95, line 1: normally we are making the solution an executable too, even if it is not necessary since we use "bash", but since it was like this for the Go piscine we also kept this process.

MSilva95 (changes requested):
The test and solution are working well!
Just needs some more guidance for the students

subjects/devops/check-user/README.md
mikysett: done

subjects/devops/check-user/README.md
mikysett: done

subjects/devops/check-user/README.md
```diff
@@ -0,0 +1,39 @@
...
+- First argument different from `-e` or `-i`: `"Error: unknown flag"`, exit with `1`.
+
+### Hints
+
+> `man getent` will be a great resource to explore
```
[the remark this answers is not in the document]
mikysett: I've added more hints about `getent`, also the link in the resources have a dedicated session about it, explaining quite well how to use it.

sh/tests/check-user_test.sh
mikysett: I've applied the autoformat, but yes it seems like a convention, not changing much

MSilva95 (approved):
LGTM

PR #1710, opened 2023-01-11
DEV-3953-grades

MSilva95:


sh/tests/grades_test.sh
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env bash
```
mikysett, line 1: The test is not checking for good behavior when more or less arguments than expected are passed.

Also in the current state there is a lot of code duplication and the tests are a bit hard to read.
It would be nice to generalize `challenge` function in order to accept any input and make the `diff` without having to fork for specific case scenarios.

In case you feel it could be useful I've refactored it to have an idea on what the shape could be:
```sh
#!/usr/bin/env bash

IFS='
'
script_dirS=$(cd -P "$(dirname "$BASH_SOURCE")" &>/dev/null && pwd)

challenge() {
    args=${@:1:$#-1}
    input=${@: -1}

    submitted=$(
        bash -c ""$script_dirS"/student/grades.sh $args" <<EOF
$input
EOF
    )
    expected=$(
        bash -c ""$script_dirS"/solutions/grades.sh $args" <<EOF
$input
EOF
    )
}

challenge 3 "Bob
90
Alice
150
Eve
55
"

challenge 3 "Bob
90
Alice
100
Eve
55
"

challenge ""

challenge 1 121 21 3 "Alice
1
"
```

Of course you will still need to add the other tests I didn't include into the example.

subjects/devops/grades/README.md
```diff
@@ -0,0 +1,108 @@
+## grades
+
+### Instructions
+
+Create a script `grades.sh` which will ask the user for a specific number of students in order to evaluate them. After the number is set, the script will ask you to introduce the "names" and "grades" of the persons you wish to evaluate.
```
mikysett, line 5: As phrased here it seems like the script will prompt to ask the number of users while it takes it as an argument instead. What do you think about rephrasing maybe making it more clear that the number of students will be passed as an argument to the script?

subjects/devops/grades/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+The grades will have a range between 0 and 100 and will be numeric values only.
+
+According to each student grade you will have to return the following:
```
mikysett, line 9: The order here should be swapped, if you start by `< 50`, `>= 50`, `>= 70` and so on by logic a student having `100` should print all those messages: `>= 50`, `>= 70`, `>= 90`.
Instead we want to print only the `>= 90`, so the swapping.

subjects/devops/grades/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+- If the student grade is anything lower than 50 you will return the string "<name>: You had a poor performance!":
```
mikysett, line 11: For consistency maybe we can wrap the expected return in back tick.
"\<name>: You had a poor performance!" -> `"<name>: You had a poor performance!"`

to do for all messages

subjects/devops/grades/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+In bash, you can use the `read` command to read input from the user and store it in a variable like so:
+
+`read var_name`
+
+You can also use the "-p" option to specify a prompt that will be displayed to the user before reading the input.
+
+Use loops to go through each student. For example the for loop is very helpful.
+
+```console
```
mikysett, line 92: maybe here it is better to have bash or sh instead of console (in the platform as it is right now it will have no effect, but if in future we will add colors then `console` won't be the right one for this code).

subjects/devops/grades/README.md
```diff
@@ -0,0 +1,108 @@
...
+Albert: You did an excellent job!
+```
+
+### Hints
+
+In bash, you can use the `read` command to read input from the user and store it in a variable like so:
+
+`read var_name`
+
+You can also use the "-p" option to specify a prompt that will be displayed to the user before reading the input.
```
mikysett, line 88: "-p" -> `-p`

subjects/devops/grades/README.md
```diff
@@ -0,0 +1,108 @@
...
+Use loops to go through each student. For example the for loop is very helpful.
+
+```console
+for element in sequence
+do
+  # commands to be executed
+done
+```
+
+You can use an array to store multiple pieces of data. To create an array, you can use the `declare` command with the "-a" option:
```
mikysett, line 99: "-a" -> `-a`

subjects/devops/grades/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+`declare -a name_array`
+
+To add elements to the array, you can use the += operator:
```
mikysett, line 103: += -> `+=`

subjects/devops/grades/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Error handling
+
+All errors will print a specific message on **stderr** (ending with a newline) and returns a specific non-zero value:
+
+- Wrong number of arguments: `"Error: expect 2 arguments"`, exit with `1`.
+
+- If the student grade is not a number or is greater than 100: `Error: The grade "grade" is not a valid input. Only numerical grades between 0 and 100 are accepted.`, exit with `1`.
+
+```console
```
mikysett, line 55: maybe we can add also an example with wrong number of arguments here

subjects/devops/grades/README.md
```diff
@@ -0,0 +1,108 @@
...
+Student Name #1: Sara
+Student Grade #1: 90
+Sara: You did an excellent job!
+```
+
+### Error handling
+
+All errors will print a specific message on **stderr** (ending with a newline) and returns a specific non-zero value:
+
+- Wrong number of arguments: `"Error: expect 2 arguments"`, exit with `1`.
```
mikysett, line 51: Actually the script expect only one argument, not two

mikysett (changes requested):
One case is not tested and not implemented in the solution
Small possible improvements in subject style
The error for wrong number arguments have a little mistake in it

subjects/devops/grades/README.md
MSilva95: done

subjects/devops/grades/README.md
MSilva95: done

subjects/devops/grades/README.md
MSilva95: done

subjects/devops/grades/README.md
MSilva95: done

subjects/devops/grades/README.md
MSilva95: done

subjects/devops/grades/README.md
MSilva95: done

subjects/devops/grades/README.md
MSilva95: done

sh/tests/solutions/grades.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
```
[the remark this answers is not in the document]
MSilva95: done

sh/tests/grades_test.sh
MSilva95: done

sh/tests/grades_test.sh
```diff
@@ -5,65 +5,50 @@ IFS='
...
-90
-Alice
-75
-Eve
-55
+        bash -c ""$script_dirS"/solutions/grades.sh $args" <<EOF
+$input
 EOF
     )
     diff <(echo "$submitted") <(echo "$expected")
```
mikysett, line 21: The tester returns a 0 even when the tests fail, it should return `1` or anything different than `0` instead.
You could add this just after the `diff`:
```bash
    if [ $? != 0 ]; then
        exit 1;
    fi
```
It will check if `diff` exited with a non 0 value, if so it means the two entries are different and the tester will `exit 1`.

sh/tests/grades_test.sh
```diff
@@ -73,32 +58,11 @@ john
...
-150
-Eve
-55
-EOF
-    )
-    diff <(echo "$submitted") <(echo "$expected")
-}
-
-challenge
+"
```
mikysett, line 68: - `0` as an imput for grade is not tested, it could be nice to add this case.
- also negative inputs are not tested.
- Input which is not a number for grade is also not tested

mikysett (changes requested):
- some style improvements
- tester not returning the right value
- possibility to add more tests

subjects/devops/grades/README.md
```diff
@@ -0,0 +1,113 @@
...
+According to each student grade you will have to return the following:
+
+- If the student grade is anything equal or greater than 90 you will return the string `"<name>: You did an excellent job!"`:
+
+```console
+$ ./grades.sh 1
+Student Name #1: Sara
+Student Grade #1: 90
+Sara: You did an excellent job!
+```
```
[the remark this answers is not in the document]
MSilva95: done

subjects/devops/grades/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+- If the student grade is anything equal or greater than 70 you will return the string `"<name>: You did a good job!"`:
+
+```console
+$ ./grades.sh 1
+Student Name #1: Sara
+Student Grade #1: 75
+Sara: You did a good job!
+```
```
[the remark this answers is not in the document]
MSilva95: done

subjects/devops/grades/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+- If the student grade is anything equal or greater than 50 you will return the string `"<name>: You need a bit more effort!"`:
+
+```console
+$ ./grades.sh 1
+Student Name #1: Sara
+Student Grade #1: 51
+Sara: You need a bit more effort!
+```
```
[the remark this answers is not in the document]
MSilva95: done

subjects/devops/grades/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+- If the student grade is anything lower than 50 you will return the string `"<name>: You had a poor performance!"`:
+
+```console
+$ ./grades.sh 1
+Student Name #1: Sara
+Student Grade #1: 34
+Sara: You had a poor performance!
+```
```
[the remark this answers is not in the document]
MSilva95: done

subjects/devops/grades/README.md
```diff
@@ -0,0 +1,113 @@
...
+- Wrong number of arguments: `"Error: expect 1 argument only!"`, exit with `1`.
+
+- If the student grade is not a number or is greater than 100: `Error: The grade "grade" is not a valid input. Only numerical grades between 0 and 100 are accepted.`, exit with `1`.
+
+```console
+$ ./grades.sh 2
+Student Name #1: Sara
+Student Grade #1: 101
+Error: The grade '101' is not a valid input. Only numerical grades between 0 and 100 are accepted.
+./grades.sh 2
```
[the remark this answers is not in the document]
MSilva95: done

subjects/devops/grades/README.md
```diff
@@ -0,0 +1,113 @@
...
+Student Grade #2: 56
+Student Name #3: James
+Student Grade #3: 78
+Student Name #4: Albert
+Student Grade #4: 90
+Sara: You had a poor performance!
+Norman: You need a bit more effort!
+James: You did a good job!
+Albert: You did an excellent job!
+```
```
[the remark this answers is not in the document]
MSilva95: done

sh/tests/grades_test.sh
MSilva95: done

sh/tests/grades_test.sh
```diff
@@ -19,6 +19,12 @@ $input
 EOF
     )
     diff <(echo "$submitted") <(echo "$expected")
+
+    if [ $? != 0 ]; then
+        exit 1
+    fi
+    echo $submitted
+    echo $expected
```
mikysett, line 27: Those two `echo` could be removed, I guess they are a reminiscence from an earlier testing stage.

sh/tests/grades_test.sh
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env bash
+
+IFS='
+'
+script_dirS=$(cd -P "$(dirname "$BASH_SOURCE")" &>/dev/null && pwd)
+
+challenge() {
+    args=$1
+    input=$2
```
mikysett, line 9: taking `args` and `input` like that doesn't allow you to test behavior when more or less than one argument are passed to `grades.sh`.

You could use this kind of notation to achieve that goal:
```bash
args=${@:1:$#-1}
input=${@: -1}
```

This is a bit of a bashinsm, and could have some portability draw-backs if used with different terminals, but here it should be quite safe to assume we will always run that script using bash since we have control over the execution environment.

If you find other ways to implement this please feel free to do so, as soon as it allows to test variadic arguments passed to the script.

sh/tests/grades_test.sh
```diff
@@ -0,0 +1,85 @@
...
+150
+Eve
+55
+"
+challenge 1 "Student1
+not_good
+"
+challenge 1 "Student1
+-75
+"
```
mikysett, line 85: We are still missing a test for the wrong number of arguments, it could be something like so:
```bash
challenge 1 1 "Louis
20
"
```

sh/tests/solutions/grades.sh
```diff
@@ -0,0 +1,34 @@
...
+num_students=$1
+declare -a students
+
+if [ $# -ne 1 ]; then
+    echo "Error: Expect 1 argument only!"
```
mikysett, line 7: Here the message is slightly different to the one expected as stated in the subject:
```
Error: expect 1 argument only!
```
In the subject `expect` is lowercase

mikysett (changes requested):
- arguments number edge case doesn't match the subject
- this edge case is not tested
- little fix on challenge to allow variadic arguments handling
- two `echo`s could be removed

sh/tests/grades_test.sh
MSilva95: done

sh/tests/grades_test.sh
MSilva95: done, it was not working before so i pushed with the other fix, it turns out we need the double quotes 
```
    input="${@: -1}"
    args=${@:1:$#-1}
```

sh/tests/grades_test.sh
MSilva95: done

sh/tests/solutions/grades.sh
MSilva95: done

mikysett (approved):
LGTM

PR #1719, opened 2023-01-17
DEV-3952 joker-num

mikysett:
New exercise for scripting piscine, devops branch

### DIFF
--- a/sh/tests/joker-num_test.sh
+++ b/sh/tests/joker-num_test.sh
@@ -1,51 +1,76 @@
-test_empty_input() {
-    output=$(echo -n "" | ./solutions/joker-num.sh)
-    expected_output="Error: Input is empty, please try again."
-    diff <(echo "$output") <(echo "$expected_output")
+#!/usr/bin/env bash
+
+IFS='
+'
+script_dirS=$(cd -P "$(dirname "$BASH_SOURCE")" &>/dev/null && pwd)
+
+challenge() {
+    args=${@:1:$#-1}
+    input="${@: -1}"
+
+    submitted=$(
+        ./student/joker-num.sh $args <<EOF
+$input
+EOF
+    )
+    expected=$(
+        ./solutions/joker-num.sh $args <<EOF
+$input
+EOF
+    )
+    diff <(echo "$submitted") <(echo "$expected")
 }
 
-test_not_a_number() {
-    output=$(echo "a" | ./solutions/joker-num.sh)
-    expected_output="Error: Input is not a number, please try again."
-    diff <(echo "$output") <(echo "$expected_output")
-}
+# Good input, win
+input="1
+100
+49
+51
+50
+"
 
-test_number_out_of_range() {
-    output=$(echo "100001" | ./solutions/joker-num.sh)
-    expected_output="Error: Number out of range, please try again."
-    diff <(echo "$output") <(echo "$expected_output")
-}
+# Good input, win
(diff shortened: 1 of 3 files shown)


subjects/devops/joker-num/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+- `read`: this command can be used to read input from the command line.
+
+For example:
+
+```console
+$ read guess
+```
+
+After running that line, you would type your password, press `Enter`, and it would be stored in the `$password` variable for later use.
```
nprimo, line 77: after running `prettier` I got this diff:

```
diff --git a/subjects/devops/joker-num/README.md b/subjects/devops/joker-num/README.md
index d2053088..0cfc2192 100644
--- a/subjects/devops/joker-num/README.md
+++ b/subjects/devops/joker-num/README.md
@@ -7,6 +7,7 @@ The script will receive the number to guess as argument. The number should be be
 To achieve this you will have to use a `for` loop.

 When guessing a valid number the output will be:
+
 - Number bigger than the secret one: `Go down`.
 - Number smaller than the secret one: `Go up`.
 - Number equal than the secret one: `Congratulations, you found the number!`.
@@ -38,6 +39,7 @@ $
 - The player writing something that is not a valid number as an answer: The program will ask again to enter a guess and the number of tries won't be decremented.

 Here an example for the last scenario:
+
 ```console
 $ ./joker-num.sh 100
 Enter your guess (5 tries left):

```

sh/tests/solutions/joker-num.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
+if [[ $# != 1
+    || -z "$1"
```
nprimo, line 4: Nice touch! Even if it's not needed to pass this exercise, could be nice to add this to the `hints` part of the subject

sh/tests/joker-num_test.sh
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env bash
+
```
nprimo, line 2: To make sure the tester exit at the first failed test, add the following:
```bash
set -e 
```

the `-o pipefail` should not add any value in this case as the `-u` for undefined variables

nprimo (changes requested):
- Check `prettier` in the subject 
- One small change for the tests 
- A suggestion for a potential additional `hint` to add in the subject

subjects/devops/joker-num/README.md
mikysett: done

sh/tests/solutions/joker-num.sh
mikysett: done

sh/tests/joker-num_test.sh
mikysett: I've added the full unofficial strict mode syntax

sh/tests/joker-num_test.sh
```diff
@@ -0,0 +1,77 @@
...
+    expected=$(
+        ./solutions/joker-num.sh $args <<EOF
+$input
+EOF
+    )
+    diff <(echo "$submitted") <(echo "$expected")
+}
+
+# Good input, win
+input="1
```
nprimo, line 26: this variable is not used

mikysett: removed

PR #1729, opened 2023-01-18
DEV-4284 replace solution in football tournament

eslopfer:
### Why?
The previous solution wasn't accurate.
### Solution Overview
Replace it for the right one.
### Implementation Details
Not relevant.


davhojt (approved):
LGTM.

PR #1730, opened 2023-01-18
DEV-4286: add tree to the dockerfile

MSilva95:

### Why?

To fix this error in the tests:
`file-struct_test.sh: line 11: tree: command not found`

### Solution Overview

Adding the tree command in the Dockerfile to fix a bug with the scripting tests

### Implementation Details


davhojt (approved):
LGTM.

PR #1732, opened 2023-01-18
Update README.md set-env-vars typo

lotoussa:


davhojt (approved):

davhojt:
@lotoussa You can actually assign me as the reviewer, instead of writing that I am the assigned person. That way, I will see it in the list of outstanding reviews.

image

lotoussa:
sure ty for the tips but i don't have the possibility to do that i think
it's like there is no button even before i create the PR.
image

PR #1736, opened 2023-01-19
class to classe

sagarishere:
class to classe as in the attached js object, there is no class tag, only classe tag mentioned. This confuses a lot of non-French speakers.

### Why?

> Classe tag is present as a classifier rather than the class tag. Non-french speakers (including me) waste disproportionate amount of time to know this mistake.

### Solution Overview

> Changed class to classe in the task description

### Implementation Details

> The alternative was to change the tag itself. As that would make many lines changed, I leave it to the wisdom of the merger-reviewer. If they want, they can reject this pull request and use the alternate method, i.e. correct the tag name from classe to class. If that method is not used, then we need to correct the task description which this request does.

### DIFF
--- a/subjects/get-them-all/README.md
+++ b/subjects/get-them-all/README.md
@@ -19,7 +19,7 @@ Complete the body of the following functions. The first three functions return a
   - 2nd array: all the non-architects.
 
 - `getClassical`:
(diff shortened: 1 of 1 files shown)


sagarishere:
classe was changed to class in the script inside the body.

PR #1744, opened 2023-01-23
DEV-4181 concat string ytrack

eslopfer:
Exercise's description.


subjects/devops/concat-string/README.md
```diff
@@ -0,0 +1,29 @@
+## concat_string
```
nprimo, line 1: The tile here is `snake_case` while the folder is `kebab-case` - to avoid confusion I would keep the same

nprimo (commented):
I left a comment about the directory name and title of the subject - we can better discuss the issue on Discord if needed :)

subjects/devops/concat-string/README.md
```diff
@@ -0,0 +1,29 @@
...
+### Instructions
+
+Create a file `concat_string.py` that contains a function `concat` which takes in two strings as parameters and performs the following operations:
+
+- Concatenates the two strings by adding a comma and space between them.
+- Returns the resulting string.
+
+### Usage
+
+Here is an example of how to use the `concat` function:
```
nprimo, line 12: In other exercises, it has been used a slightly different approach. 

The python code is presented as the content of a script `test.py`. It is then reported the behaviour of the script as follows:
```console
$ python3 test.py
<expected output>
$
```

eslopfer: Done. Thanks

subjects/devops/concat-string/README.md
eslopfer: Done. Thanks.

subjects/devops/concat_string/README.md
```diff
@@ -0,0 +1,31 @@
...
+string1 = "Hello"
+string2 = "World"
+print(concat_string.concat(string1, string2))
+```
+
+This will output:
+
+```console
+$ python3 test.py
+Hello World
```
nprimo, line 25: According to the subject and solution provided, the output should be:

```console
Hello, World
```

nprimo (changes requested):
A small discrepancy in the output presented +
2 suggestions

subjects/devops/concat_string/README.md
eslopfer: Good catch, thanks.

nprimo (approved):
LGTM!

PR #1746, opened 2023-01-23
DEV-4048 change name of bloody sunday exercise

eslopfer:

### Why?

### Solution Overview

### Implementation Details


eslopfer:
> LGTM.
> 
> Do you have a PR for the `/all` repo too? Because some exercises will be pointing at the bloody sunday exercise - but won't find the subject.
> 
> Let's not merge this to master until next Monday, so that we change the path to the subject closer to the time that all will be redeployed.

I made the change in the refs and assigned you as reviewer now. Thanks :+1:

PR #1747, opened 2023-01-23
DEV-4307 docs(invert): clarify what to return

eslopfer:
### Why?

Inaccurate description

### Solution Overview
Specify what to return

### Implementation Details
Doesn't apply


davhojt (approved):
LGTM.

PR #1749, opened 2023-01-23
Dev 4182 school punishment

mikysett:
Add a new python exercise to scripting piscine


subjects/devops/punishment/README.md
```diff
@@ -0,0 +1,57 @@
...
+```
+
+### Hints
+
+- Removing spaces at the start and end of a string is so common that almost all languages implement this feature. Here you can check for the `strip()` method.
+- Instead of using loops you can try the `string multiplication` operator, which is a very nice feature of Python and will make your code more readable.
+
+### References
+
+- [strip](https://www.w3schools.com/python/ref_string_strip.asp)
```
nprimo, line 56: Nothing major, but I would use the official python doc https://docs.python.org/3.11/library/stdtypes.html?highlight=strip#str.strip

nprimo (approved):
I left a small suggestion, but I feel the subject is good to go. However, the automated checks are not passing for some reasons

subjects/devops/punishment/README.md
mikysett: done

PR #1752, opened 2023-01-23
DEV-4180 numerical-operations

mikysett:
add new python exercise to scripting piscine


MSilva95 (changes requested):
I've added a commit to add `apt-utils` in the dockerfile so that the pr passed all actions in public.

MSilva95 (approved):
LGTM!

PR #1762, opened 2023-01-24
DEV 4184 - shopping exercise

nprimo:
New exercise for scripting piscine quest 6 - subject

### DIFF
--- a/subjects/devops/shopping/README.md
+++ b/subjects/devops/shopping/README.md
@@ -1,0 +1,41 @@
+# shopping
+
+## Instructions
+
+Create a file `shopping.py` with a function `remember_the_apple(shopping_list)`.
+
+This function input is a list of string, like this:
+
+```python
+['tomatoes', 'pastas', 'apple', 'salt']
+```
+
+If the string `'apple'` is not in the list, it should be added. The function return the updated list. If the apple is in the list, simply return it.
+
+If the input list is empty, the returned list should be also empty (don't add the apple in that case).
+
+## Usage
+
+Here is a possible `test.py` to test your functions:
+
+```python
+import shopping
+
+if __name__ == '__main__':
+    list_with_apple = ['tomatoes', 'pastas', 'apple', 'salt']
+    list_without_apple = ['tomatoes', 'pastas', 'salt']
+
+    print(shopping.remember_the_apple(list_with_apple))
+    print(shopping.remember_the_apple(list_without_apple))
+```
+
+```bash
+$ python test.py
+['tomatoes', 'pastas', 'apple', 'salt']
+['tomatoes', 'pastas', 'salt', 'apple']
+```
+
+## Notions
+
+- [lists](https://docs.python.org/3/tutorial/datastructures.html#)
+- [Conditions](https://docs.python.org/3/tutorial/datastructures.html#more-on-conditions)
--- a/subjects/devops/shopping/README.md
+++ b/subjects/devops/shopping/README.md
@@ -1,10 +1,10 @@
-# shopping
+## shopping
 
-## Instructions
+### Instructions
 
 Create a file `shopping.py` with a function `remember_the_apple(shopping_list)`.
 
-This function input is a list of string, like this:
+This function input is a list of strings, like this:
 
 ```python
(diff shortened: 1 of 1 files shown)


subjects/devops/shopping/README.md
```diff
@@ -0,0 +1,41 @@
+# shopping
+
+## Instructions
```
mikysett, line 3: it should be an H3 `###`

subjects/devops/shopping/README.md
```diff
@@ -0,0 +1,41 @@
+# shopping
```
mikysett, line 1: it should be an H2 `##`

subjects/devops/shopping/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+```python
+['tomatoes', 'pastas', 'apple', 'salt']
+```
+
+If the string `'apple'` is not in the list, it should be added. The function return the updated list. If the apple is in the list, simply return it.
+
+If the input list is empty, the returned list should be also empty (don't add the apple in that case).
+
+## Usage
```
mikysett, line 17: it should be an H3 `###`

subjects/devops/shopping/README.md
```diff
@@ -0,0 +1,41 @@
...
+    print(shopping.remember_the_apple(list_without_apple))
+```
+
+```bash
+$ python test.py
+['tomatoes', 'pastas', 'apple', 'salt']
+['tomatoes', 'pastas', 'salt', 'apple']
+```
+
+## Notions
```
mikysett, line 38: it should be an H3 `###`

subjects/devops/shopping/README.md
```diff
@@ -0,0 +1,41 @@
+# shopping
+
+## Instructions
+
+Create a file `shopping.py` with a function `remember_the_apple(shopping_list)`.
+
+This function input is a list of string, like this:
```
mikysett, line 7: I think "string" -> "strings"

subjects/devops/shopping/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+    print(shopping.remember_the_apple(list_with_apple))
+    print(shopping.remember_the_apple(list_without_apple))
+```
+
+```bash
+$ python test.py
+['tomatoes', 'pastas', 'apple', 'salt']
+['tomatoes', 'pastas', 'salt', 'apple']
+```
```
mikysett, line 36: Usually we add a line with `$` at the end to explicitly say the program ended and terminal is back

subjects/devops/shopping/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Notions
+
+- [lists](https://docs.python.org/3/tutorial/datastructures.html#)
```
mikysett, line 40: "lists" -> "Lists"

subjects/devops/shopping/README.md
```diff
@@ -0,0 +1,41 @@
...
+If the input list is empty, the returned list should be also empty (don't add the apple in that case).
+
+## Usage
+
+Here is a possible `test.py` to test your functions:
+
+```python
+import shopping
+
+if __name__ == '__main__':
```
mikysett, line 24: We don't have a convention, but after I documented a bit I decided not to include that in my examples, since `test.py` is meant to be executed as a script anyway and having this conditionally executed code is useless, since it will make sense only if `test.py` will be imported as a library by another script, which actually makes no sense.

Still, not having a convention fell free to ignore this comment if you think this line should stay, but I'll be of course very happy if you can share your point of view about it ;)

mikysett (changes requested):
small styling issues to fix, everything else seems quite cool!

subjects/devops/shopping/README.md
nprimo: Totally correct! Thank you for the reminder

subjects/devops/shopping/README.md
nprimo: I put it "out of habit" even if I haven't coded with python for a while - I will remove it also not to confuse the student with something that might look weird to someone that is not familiar with the language.

mikysett (approved):
LGTM

PR #1763, opened 2023-01-24
DEV-4318 merge-two

mikysett:
add new python exercise to scripting piscine

subject and tests/solution should be merged at the same time

### DIFF
--- a/subjects/devops/merge_two/README.md
+++ b/subjects/devops/merge_two/README.md
@@ -1,0 +1,53 @@
+## Merge two
+
+One very useful data structure in Python are the dictionaries, in this exercise we will start to familiarize and use them.
+
+### Instructions
+
+Create a file `merge_two.py` which will have a function named `merge_two()`. This function will accept two dictionaries and return a third one which will be the merge of the two inputs.
+
+Here is the prototype of the function:
+
+```python
+def merge_two(first_dict, second_dict):
+    # this is a function,
+    # write your code here
+```
+
+### Usage
+
+Here is a possible code to test your function. Put it in another file (ex: `test.py`):
+
+```python
+from merge_two import merge_two
+
+first = {
+    "Bob": 36,
+    "Louise": 23,
+    "Lea": 34
+}
+second = {
+    "Louise": 44,
+    "Romolo": 30,
+    "Lea": 22
+}
+print(merge_two(first, second))
+```
+
+Run your test file with the following command:
+
+```console
+$ python3 test.py
+{'Bob': 36, 'Louise': 44, 'Lea': 22, 'Romolo': 30}
+$
+```
+
+### Hints
+
+- If a key is repeated in both dictionaries the value retained will be the one in `second_dict`.
+
+- There is different ways to merge dictionaries, take the time to understand the differences in between those techniques and try more than one technique to better retain it.
+
+### References
+
+- [Merging dictionaries in Python](https://www.geeksforgeeks.org/python-merging-two-dictionaries/)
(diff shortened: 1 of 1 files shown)


subjects/devops/merge_two/README.md
```diff
@@ -0,0 +1,53 @@
...
+$ python3 test.py
+{'Bob': 36, 'Louise': 44, 'Lea': 22, 'Romolo': 30}
+$
+```
+
+### Hints
+
+- If a key is repeated in both dictionaries the value retained will be the one in `second_dict`.
+
+- There is different ways to merge dictionaries, take the time to understand the differences in between those techniques and try more than one technique to better retain it.
```
nprimo, line 49: It should be 
```
- There are different ways ... 
```

nprimo (commented):
In the JIRA ticker it is mentioned to convert the dictionary to JSON - is there a particular reason this has not been introduced into the subject?

subjects/devops/merge_two/README.md
mikysett: done

mikysett:
Thanks a lot for pointing out `JSON` features are missing.
I've implemented it now, also `stdin` input was supposed to be implemented but I think it will be way too much work to properly implement a test for this situation. I've started a thread with @zamazzal to see what he thinks could be the best approach

subjects/devops/merge_two/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+- Use the function `input()` to read from `stdin` and `int()` to convert the value to a number.
+
+> Your solution will be tested only for valid inputs (all the values will be convertible to `int`).
+
+### References
+
+- [Merging dictionaries in Python](https://www.geeksforgeeks.org/python-merging-two-dictionaries/)
+- [JSON library in Python](https://docs.python.org/3/library/json.html)
+- [Function input() in Python](https://www.w3schools.com/python/ref_func_input.asp)
```
nprimo, line 73: I get this after running `prettier`

```console
diff --git a/subjects/devops/merge_two/README.md b/subjects/devops/merge_two/README.md
index e8feea74..21ff1da9 100644
--- a/subjects/devops/merge_two/README.md
+++ b/subjects/devops/merge_two/README.md
@@ -70,4 +70,4 @@ $

 - [Merging dictionaries in Python](https://www.geeksforgeeks.org/python-merging-two-dictionaries/)
 - [JSON library in Python](https://docs.python.org/3/library/json.html)
-- [Function input() in Python](https://www.w3schools.com/python/ref_func_input.asp)
\ No newline at end of file
+- [Function input() in Python](https://www.w3schools.com/python/ref_func_input.asp)
(END)
```

subjects/devops/merge_two/README.md
```diff
@@ -0,0 +1,73 @@
+## Merge two
+
+One very useful data structure in Python are the dictionaries, in this exercise we will start to familiarize and use them.
+
+### Instructions
+
+Create a file `merge_two.py` which will have a function named `merge_two()`. This function will accept one dictionary.
+It will prompt the user to create a new dictionary asking for keys and values.
```
nprimo, line 8: Maybe add also something like: 
```
The function will always try to convert the `values` to integer
```

nprimo (changes requested):
I left 2 comments inline

subjects/devops/merge_two/README.md
mikysett: done

subjects/devops/merge_two/README.md
mikysett: done

nprimo (approved):
LGTM :rocket:

PR #1772, opened 2023-01-26
DEV-4326: write_file

MSilva95:

### Why?

### Solution Overview

### Implementation Details

### DIFF
--- a/subjects/devops/write_file/README.md
+++ b/subjects/devops/write_file/README.md
@@ -1,6 +1,6 @@
-# Write file
+## Write file
 
-## Instructions
+### Instructions
 
 Create a file `write_file.py` which will have a function `to_do(input)` with one input a `list[tuple[dt.date, str]]` which will be a to do list example:
 
@@ -22,7 +22,7 @@ $
 
 Make sure you follow the correct format for the output.
 
-## Usage
+### Usage
 
 Here is a possible `test.py` to test your function:
 
@@ -45,7 +45,7 @@ Wednesday 01 June 2022: fix the seat
 Thursday 02 June 2022: take the trash
 ```
 
-## Hints
+### Hints
 
 - The `strftime() `method is used to format the date object as a string. You can use this method to format the date in the desired format (e.g. %A %d %B %Y).
 
@@ -80,7 +80,7 @@ The `write()` method is used to write the string "Hello World!" to the file.
 
 - Test your function with different input formats, different date and task length, different date range and make sure that it works as expected.
 
-## References
+### References
 
 - [open a file](https://www.w3schools.com/python/python_file_write.asp)
 - [write in a file](https://www.pythontutorial.net/python-basics/python-write-text-file/)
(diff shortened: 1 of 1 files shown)


subjects/devops/write_file/README.md
```diff
@@ -0,0 +1,86 @@
+## Write file
+
+### Instructions
+
+Create a file `write_file.py` which will have a function `to_do(input)` with one input a `list[tuple[dt.date, str]]` which will be a to do list example:
```
nprimo, line 5: Nothing major - my spell checker suggest to spell to do list as follows: `to-do list`

subjects/devops/write_file/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+```python
+[
+    (dt.date(2022, 6, 1), "fix the seat"),
+    (dt.date(2022, 6, 2), "take the trash"),
+]
+```
+
+Your function needs to format and write the input in a text file `output.txt` as follow:
```
nprimo, line 14: `as follow` should be `as follows`

subjects/devops/write_file/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+The `write()` method is used to write the string "Hello World!" to the file.
+
+- Make sure to include a newline character `\n` at the end of each line of output, to separate the tasks in the output file.
+
+- Test your function with different input formats, different date and task length, different date range and make sure that it works as expected.
+
+### References
+
+- [open a file](https://www.w3schools.com/python/python_file_write.asp)
```
nprimo, line 85: Just a suggestion: I would give a link that showcases the `with open('file.txt') as file: ...` construct. For example, this one https://docs.python.org/3/tutorial/inputoutput.html#tut-files

subjects/devops/write_file/README.md
```diff
@@ -0,0 +1,86 @@
...
+Your function needs to format and write the input in a text file `output.txt` as follow:
+
+```bash
+$ cat output.txt
+Wednesday 01 June 2022: fix the seat
+Thursday 02 June 2022: take the trash
+$
+```
+
+Make sure you follow the correct format for the output.
```
nprimo, line 23: What should happen when the list passed is empty?

What should happen if the format of the list is not the one required? Should the student always assume that the list is correctly formatted?

nprimo (changes requested):
I left some suggestions inline

subjects/devops/write_file/README.md
MSilva95: done

subjects/devops/write_file/README.md
MSilva95: done

subjects/devops/write_file/README.md
MSilva95: Done, thanks for the link!

subjects/devops/write_file/README.md
MSilva95: i will add clarification on the readme, but yes they should assume the list is correct.

nprimo (approved):
LGTM! :rocket:

PR #1777, opened 2023-01-31
DEV-4327 - skip secrete new python exercise

nprimo:
Add new python scripting piscine exercise


mikysett (approved):
LGTM

PS: maybe it's better to wait for solution/tests to be approved also before to merge

PR #1779, opened 2023-02-01
Fix typo in keras (no issue id)

Whovillage:

### Why?

The exercise suggest we use categorical_crossentropy , but the classes in breast cancer data set are actually binary, so binary cross entropy should be used to get the audit answers. 

### Solution Overview

change the wording from latter to former


eslopfer (dismissed):
Good catch! 

Thanks for your contribution.

LGTM :rocket:

PR #1780, opened 2023-02-01
DEV-4373 add missing exercise to the description

eslopfer:
In the quest keras2, when merging all the exercises into one, one was left out.

I went to the original repository to get it back.


davhojt (approved):
LGTM.

PR #1789, opened 2023-02-07
DEV-4377-q-1-optional-largest

nprimo:
- add subject, solution, test and support files


subjects/devops/largest/README.md
```diff
@@ -0,0 +1,46 @@
...
+    5 | ./f1$
+    3 | ./d1/f2-1$
+   38 | ./d1/f2-3$
+   34 | ./d1/f2-2$
+   34 | ./f3$
+    2 | ./d1/f1-1$
+$
+```
+
+> The output needs to be formatted as the example shown above
```
mikysett, line 42: Maybe for consistency let's add a `.` at the end of the sentence

subjects/devops/largest/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+### Hints
+
+- `awk` might be handy
```
mikysett, line 46: Maybe for consistency let's add a `.` at the end of the sentence

mikysett (changes requested):
There is 2 small style improvements.

Also I would strongly suggest to remove all the files and directories in `largest-dir` (including largest-dir itself) from the PR.
Instead I think it would be cleaner to programmatically create those dirs and files as a prelude in `largest_test.sh` using `mkdir` and `touch`... If I'm not wrong touch also have a flag to specify that it could and should create the parent dirs if needed

subjects/devops/largest/README.md
eslopfer: Thanks. Done

subjects/devops/largest/README.md
eslopfer: Thanks. Done.

mikysett (approved):
LGTM!

PR #1791, opened 2023-02-08
feat(backup_manager): add new exercise subject

mikysett:
Quest 8 - Scripting piscine


MSilva95 (commented):
Is there any reason to remove all the other information?
It's just that we were instructed to give as much help as possible in the readme that's why I had put so many details and examples.
But do you think this is enough for someone who has no knowledge to complete the project in two days?

mikysett:
As we spoke I've moved the logging info at the top of the subject and added extensive examples of logging messages in the list of actions of `backup_manager.py`. Like so it should be way clearer to the student how to organize the scripts in order to log in a correct way every action taken by the two scripts.

MSilva95 (approved):
LGTM!

PR #1792, opened 2023-02-08
DEV-4412 docs(make-your-game): add missing question

eslopfer:
There was an instruction that was making the next instruction confusing, so I added the question that seemed to be missing between both.


davhojt (approved):
LGTM.

PR #1795, opened 2023-02-10
DEV-4422-add-pictures-to-deep-in-system-repo

zamazzal:

### Why?

avoid using pictures in assets

### Solution Overview

add pictures to the repo instead of assets

### Implementation Details

add pictures to the repo instead of assets and modify the readme file


kigiri (approved):
Not really worth it in that case but don't hesitate to compresse images with tools like https://squoosh.app/ before adding them to repositories

PR #1796, opened 2023-02-10
DEV-4430 Add `runner` v2 support

HarryVasanth:
### Why?

Currently, the automated deployments rely on the `runner` service to run with the self-hosted docker images (`docker.01-edu.org`). However, recently we implemented GitHub based images in the form of GitHub packages. These packages needs to be retrieved by the `runner` for running newer tests.

### Solution Overview

Add `runner` v2 support for the configure.sh script. Which includes the deployment capabilities for the `runner` to run docker images from both GitHub and self-hosted docker images (`docker.01-edu.org`).

### Implementation Details

The `runner` has already been modified to accommodate these changes. The automated deployment script now takes the additional parameters during the execution to deploy the `runner` accordingly.

### DIFF
--- a/sh/debian/configure.sh
+++ b/sh/debian/configure.sh
@@ -258,9 +258,23 @@ function deployCore() {
   # Deploy Runner
   echo -e "Deploying Runner service: \n "
   cd /root/core/runner
+
+  # Get the latest release version tag and create latest branch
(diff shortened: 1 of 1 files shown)


MSilva95 (approved):
LGTM!

PR #1801, opened 2023-02-13
crossword audit

taivox:
Fixed audit questions for crossword.


MSilva95:
will be closed because it is done in this one #1818

PR #1806, opened 2023-02-14
Fix and improve crossword audit

### LINKED ISSUE #1805: crossword subject
I have several audit fixes and one proposal:
1. There's an audit question with an invalid comment. `Test multiple solutions` is a `Test no solutions`
2. Tests are not complete. There's no multiple solutions case (which should be handled as an Error)
3. There are no tests for backtracking. In all audit questions, the input words are in the best order possible, so if backtracking (or another invalid solution recovering functionality) is invalid, it won't be handled by test cases. We can add one more audit question with reversed words from the biggest example. In this case, words will be in the worst order possible.
4. There's no time limit for tests. The solution that tries to add words to the board randomly is valid according to the audit questions, even though it takes several days to process.

5. Proposal. It's better to change expected function usage to return a solution or error instead of printing. It will make the function reusable and make it possible to implement unit tests. It's a very good task to learn unit testing in javascript (especially for performance improvements). We can add questions about tests as a bonus.

maximihajlov:
resolves #1805

### DIFF
--- a/subjects/crossword/README.md
+++ b/subjects/crossword/README.md
@@ -8,13 +8,13 @@ The function takes two arguments:
 1. an empty puzzle, passed as a string and
 2. a list of words to fill in the puzzle (no double words allowed)
 
-The function must print on the console a string representing the puzzle filled with the input words.
+The function must return a string representing the puzzle filled with the input words.
 
 The empty puzzle will be a string with the following rules:
 - each character will be either a number, a `.` or a `\n`;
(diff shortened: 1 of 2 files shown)


MSilva95 (commented):
Is there any reason to remove all the other information?
It's just that we were instructed to give as much help as possible in the readme that's why I had put so many details and examples.
But do you think this is enough for someone who has no knowledge to complete the project in two days?

MSilva95:
will be closed because of this pr #1818 
Thanks for the feedback!

Issue #1802, opened 2023-02-14
middle_day subject


codecplyre:
middle_day
---
Utc.ymd is deprecated. Are we supposed to make it work this way?

mikysett:
Hello @codecplyre 

Thanks a lot for spotting this warning, I've updated the concerned docs and code.

Have a nice day and thanks again :)

PR #1811, opened 2023-02-20
DEV 4454 fix-deprecate-method-in-middle-day-rust-piscine

mikysett:
better if `public` and `rust-tests` are merged at the same time.
[Related github issue](https://github.com/01-edu/public/issues/1802)


jotapero (approved):
LGTM :alien:

PR #1817, opened 2023-02-22
DEV-4390: new readmes for the exams exercises

MSilva95:
DEV-4390

### Why?

>  Some of the exercises are using the import "piscine" and they need to be removed on the exams.

### Solution Overview

>  For now, the fix was to create duplicated subjects and new refs for the exercises in the exam, but in the future they will to be removed

### DIFF
--- a/subjects/findprevprime/README.md
+++ b/subjects/findprevprime/README.md
@@ -21,15 +21,11 @@ Here is a possible program to test your function :
 ```go
 package main
 
-import (
-	"fmt"
-
-	"piscine"
-)
+import "fmt"
 
 func main() {
-	fmt.Println(piscine.FindPrevPrime(5))
-	fmt.Println(piscine.FindPrevPrime(4))
+	fmt.Println(FindPrevPrime(5))
+	fmt.Println(FindPrevPrime(4))
 }
 ```
 
--- a/subjects/slice/README.md
+++ b/subjects/slice/README.md
@@ -23,19 +23,15 @@ Here is a possible program to test your function :
 ```go
 package main
 
-import (
-	"fmt"
-
-	"piscine"
-)
+import "fmt"
 
 func main(){
     a := []string{"coding", "algorithm", "ascii", "package", "golang"}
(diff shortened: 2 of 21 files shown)


subjects/findprevprime/README.md
```diff
@@ -21,15 +21,11 @@ Here is a possible program to test your function :
...
 
-import (
-	"fmt"
-
-	"piscine"
-)
+import "fmt"
 
 func main() {
-	fmt.Println(piscine.FindPrevPrime(5))
```
eslopfer, line 31: May I ask why this one was modified instead of creating a new one with a new ref like the rest?

If it doesn't require any action, feel free to resolve the conversation :slightly_smiling_face:

subjects/slice/README.md
```diff
@@ -23,19 +23,15 @@ Here is a possible program to test your function :
...
+import "fmt"
 
 func main(){
     a := []string{"coding", "algorithm", "ascii", "package", "golang"}
-    fmt.Printf("%#v\n", piscine.Slice(a, 1))
-    fmt.Printf("%#v\n", piscine.Slice(a, 2, 4))
-    fmt.Printf("%#v\n", piscine.Slice(a, -3))
-    fmt.Printf("%#v\n", piscine.Slice(a, -2, -1))
-    fmt.Printf("%#v\n", piscine.Slice(a, 2, 0))
+    fmt.Printf("%#v\n", Slice(a, 1))
```
eslopfer, line 30: Like before, may I ask why this one was modified instead of creating a new one with a new ref like the rest?

If it doesn't require any action, feel free to resolve the conversation slightly_smiling_face

eslopfer (approved):
I left a couple of comments due to inconsistencies, depending on if they require action or not it looks good to be merged :rocket:

subjects/findprevprime/README.md
MSilva95: This one is only used in the exams, so it can be changed in the current readme

subjects/slice/README.md
MSilva95: same thing here

PR #1818, opened 2023-02-24
DEV-4466: crosword

MSilva95:
adding test case requested by an issue


eslopfer (approved):
Thank you for the pull request. It looks good overall, but I would like to suggest that in the future, each commit should aim to do only one thing so that they can be easily reversed if necessary. 

This makes it easier to track changes and troubleshoot issues. IMHO there are at least two commits needed for the changes in this PR, adding new test cases to the audit, and the format changes applied by the formatting tool everywhere else.

I'll leave it up to you to decide whether to make changes to the commit structure or leave it as is. 

LGTM! :rocket:

PR #1822, opened 2023-02-27
DEV-4383 - add new optional exercise (job-regist) for Q3 scripting piscine 

nprimo:


mikysett (approved):
LGTM!

The test is printing some lines that could be removed in `wait_test_case` function:
```bash
        echo -n "."
        echo $(jobs)
```

PR #1823, opened 2023-02-27
DEV-4382 q 4 optional better cat

eslopfer:
Optional exercise for the scripting piscine.


subjects/devops/better-cat/README.md
```diff
@@ -0,0 +1,64 @@
+## better-cat
+
+### Instructions
+
+Create a script `better-cat.sh` that will behave similarly to the cat command, but with additional functionality.
```
nprimo, line 5: I would highlight the `cat` - either with the ` or bold/italic - for better readibility

subjects/devops/better-cat/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+The script should accept any number of filenames as arguments, and it should print the content of each file to the standard output. If no arguments are provided, the script should print the content of all files in the current directory.
+
+The script should also support the following options:
+
```
nprimo, line 10: To increase the clarity of the subject, I would highlight the flags either with ` or with bold/italic

subjects/devops/better-cat/README.md
```diff
@@ -0,0 +1,64 @@
...
+- -r: If this option is provided, the script should print a summary line at the end of the output, indicating the total number of lines and characters printed.
+
+If both -c and -l are provided, the line number should include only non-comment lines.
+
+### Usage
+
+```console
+$ bash better-cat.sh file1.txt
+Here would be the content of file1
+
```
nprimo, line 22: Just to confirm, are these newlines intended to be here?

sh/tests/solutions/better-cat.sh
```diff
@@ -0,0 +1,51 @@
...
+  line_num=0
+  while read -r line; do
+    line_num=$((line_num+1))
+    if [[ $exclude_comments && $line == \#* ]]; then
+      continue
+    fi
+    if [[ $show_length ]]; then
+      echo "$line_num ($(( ${#line} ))): $line"
+    else
+      echo "$line_num: $line"
```
nprimo, line 37: This should not be the default behaviour (If I understood properly the subject): the current solution is always printing the number of the line, even without the `-l` flag

subjects/devops/better-cat/README.md
```diff
@@ -0,0 +1,64 @@
...
+Here would be line 10 of file3
+
+$ bash better-cat.sh -l file1.txt
+1: Here would be the content of file1
+
+$ bash better-cat.sh -r file1.txt
+Here would be the content of file1
+Total: 1 lines, 35 characters
+
+$ bash better-cat.sh -cl commented.txt
```
nprimo, line 51: Here the line should begin with `#: ` correct?

subjects/devops/better-cat/README.md
```diff
@@ -0,0 +1,64 @@
...
+$ bash better-cat.sh file1.txt file2.txt
+Here would be the content of file1
+Here would be the content of file2
+
+$ bash better-cat.sh
+Here would be the content of file1
+Here would be the content of file2
+Here would be the content of file3
+
+$ bash better-cat.sh -c commented.txt
```
nprimo, line 32: I find the below example a bit confusing: you are still referring to `file3` but have passed the argument `commented.txt`.
Additionally, which are the commented lines that should not appear - it might be enough to only skip some of the lines of the 10 that are here

subjects/devops/better-cat/README.md
```diff
@@ -0,0 +1,64 @@
...
+### Instructions
+
+Create a script `better-cat.sh` that will behave similarly to the cat command, but with additional functionality.
+
+The script should accept any number of filenames as arguments, and it should print the content of each file to the standard output. If no arguments are provided, the script should print the content of all files in the current directory.
+
+The script should also support the following options:
+
+- -c: If this option is provided, the script should exclude all lines that begin with the # character.
+- -l: If this option is provided, the script should print the line number before each line of output.
```
nprimo, line 12: Make the description of `-l` consistent with the solution implemented

subjects/devops/better-cat/README.md
```diff
@@ -0,0 +1,64 @@
...
+- -r: If this option is provided, the script should print a summary line at the end of the output, indicating the total number of lines and characters printed.
+
+If both -c and -l are provided, the line number should include only non-comment lines.
```
nprimo, line 15: I would highlight the `-c` and `-l` with ` or bold/italic

sh/tests/better-cat_test.sh
```diff
@@ -0,0 +1,41 @@
...
+}
+
+challenge "file1.txt"
+challenge "file1.txt" "file2.txt" "file3.txt"
+challenge *.txt
+challenge
+challenge -c "commented.txt"
+challenge -l "file1.txt"
+challenge -r "file1.txt"
+challenge -cl "commented.txt"
```
nprimo, line 38: I would add a case with `-clr`

sh/tests/better-cat_test.sh
```diff
@@ -0,0 +1,41 @@
...
+echo "This is the content of file1" > "file1.txt"
+echo "# commented" > "commented.txt"
+
+challenge() {
+    submitted=$(bash "$script_dir"/student/better-cat.sh "$@")
+    expected=$(bash "$script_dir"/solutions/better-cat.sh "$@")
+    diff <(echo "$submitted") <(echo "$expected") 
+}
+
+challenge "file1.txt"
```
nprimo, line 31: What if you pass a file that does not exist?

nprimo (changes requested):
Amazing solution! :rocket: I left some comments to clarify the subject and some suggestions to make the tester more well rounded

subjects/devops/better-cat/README.md
eslopfer: Done

subjects/devops/better-cat/README.md
eslopfer: Done

subjects/devops/better-cat/README.md
eslopfer: Thanks, I removed them since they could be confusing :+1:

sh/tests/solutions/better-cat.sh
eslopfer: The behaviour is correct, is the description that is not clear, the option -l should print the length according to the original specs, I am also going to modify the usage examples.

subjects/devops/better-cat/README.md
eslopfer: Yes in all cases, as it is the default behaviour, I'll modify the examples to reflect this, in this case is missing the length of the lines as well

subjects/devops/better-cat/README.md
eslopfer: Fixed. Thanks.

subjects/devops/better-cat/README.md
eslopfer: Done

subjects/devops/better-cat/README.md
eslopfer: Done.

sh/tests/better-cat_test.sh
eslopfer: Done

sh/tests/better-cat_test.sh
eslopfer: I didn't add it since the way cat behaves when a file does not exist is similar to the the default behaviour and the exercise is complex enough as it is.

I thought about adding a line clarifying that it won't be tested for inexistent files, but since in the description of the exercise that I made is not asking to handle this situation specifically I thought it would be redundant, specially considering this an optional exercise(not one of the very verbose almost step-by-step hints ones).

However if you strongly disagree let me know and I will make the changes :slightly_smiling_face:

subjects/devops/better-cat/README.md
```diff
@@ -0,0 +1,42 @@
+## better-cat
+
+### Instructions
+
+Create a script `better-cat.sh` that will behave similarly to the `cat` command, but with additional functionality.
+
+The script should accept any number of filenames as arguments, and it should print the content of each file to the standard output. If no arguments are provided, the script should print the content of all files in the current directory.
```
nprimo, line 7: It is clear from the example provided, but it might be helpful to know that the script should `... print the content of each file ...` **preceded by the line number of the file** (or something similar) `... to the standard output`

subjects/devops/better-cat/README.md
```diff
@@ -0,0 +1,42 @@
...
+- `-r`: If this option is provided, the script should print a summary line at the end of the output, indicating the total number of lines in the file and the count of characters in the lines printed.
+
+### Usage
+
+```console
+$ echo "This is an example line of text." > example.txt
+$ bash better-cat.sh example.txt
+1: This is an example line of text.
+$ bash better-cat.sh file1.txt file2.txt
+Here would be the content of file1
```
nprimo, line 22: I see what you mean here, but it might be confusing since it looks like `Here would be the content of file..` could be the actual content and it is not formatted as the default behaviour - I don't really have a clearer option to suggest but I wanted to share this thought

sh/tests/better-cat_test.sh
nprimo: I don't have a strong opinion on this: I found that it is always easier to state what should happen when someone try to use the script in a "wrong way", but at the same time we can keep it simple and add a line if someone is going to complain in the future :)

nprimo (changes requested):
Very small clarification for the subject - everything else looks good to me :rocket:

subjects/devops/better-cat/README.md
eslopfer: Thanks, you are right, this is specified in the jira ticket and it can be seen in the usage examples but is never being asked directly. I'll add it :+1:

subjects/devops/better-cat/README.md
eslopfer: Thank you for sharing your thoughts on this. I understand your concern. However, as this exercise is one of the optional advanced ones, I think it's OK to intentionally leave some room for thought. 

While I agree that it could be more explicitly described, I believe that not giving everything over-specified encourages critical thinking and problem-solving skills. Since there isn't a a clear suggestion, I will leave it as is. Thank you again for your input, and please feel free to reply to this comment and to share any other thoughts or suggestions you may have as I find them very valuable :slightly_smiling_face:

sh/tests/better-cat_test.sh
eslopfer: At this point, since the exercise description doesn't require handling inexistent files, I agree with you and I think it's reasonable to leave it as is. However, if any users have questions or concerns in the future, we can always revisit this topic and add a line to clarify the behaviour when a file doesn't exist. Thank you for your input and collaboration!

nprimo (approved):
LGTM :rocket:

PR #1829, opened 2023-02-28
DEV-4475 Renaming css files to  data.css for all dom exercises

LEEDASILVA:
DEV-4475

There was a bug when running the DOM exercises, the tests expecting the students to have the css files by doing the following : 
```js
  [`/${exercise}/${exercise}.css`]: path.join(solutionPath, `${exercise}.css`),
``` 

So by renaming the exercises it will not expect for the css files anymore


MSilva95 (approved):
LGTM!
image

PR #1832, opened 2023-02-28
DEV-4480: fix test for json-research

nprimo:
Add boilerplate to test as in `who-are-you_test.sh` that uses `curl`


davhojt (approved):
LGTM.

PR #1851, opened 2023-03-08
DEV-4446 kept promise exercise for js dev ops piscine

nprimo:


subjects/devops/kept-promise/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Example
+
+The following script `main.js` can be used to test your function:
+
+```js
+import processInfo from './kept-promise.js'
+
+const getImportantInfo = _ =>
+  new Promise(resolve => {
```
eslopfer, line 30: maybe we could add a verbose hint on how promises work? In a similar way as in the scripting piscine?

subjects/devops/kept-promise/README.md
```diff
@@ -0,0 +1,52 @@
...
+Sometimes it is important to wait for a function to finish before keep going on.
+
+Create an asynchronous function `processInfo` in a file `kept-promise.js`. The function `processInfo` will take an asynchronous function as an input, and it will print the message `Ok!` if the return of the input function is zero or a multiple of two, `Error!` otherwise.
+
+The following function can be used as an example of input function for `processInfo`:
+
+```js
+const getImportantInfo = _ =>
+  new Promise(resolve => {
+    setTimeout(() => resolve(Math.round(Math.random() * 10)), 1000)
```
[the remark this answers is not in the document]
nprimo: This is out of the scope of the solution, which is why I haven't included anything about it since it is not "required knowledge". Adding something related to `setTimeout` might be confusing in my opinion

subjects/devops/kept-promise/README.md
nprimo: Adding a `hint` on how to wait for a `Promise` would be almost like showing the solution, that's why I haven't included anything

js/tests/kept-promise_test.js
```diff
@@ -0,0 +1,19 @@
+export const tests = []
+
+const testZero = async _ => {
+  return (await processInfo(_ => new Promise(resolve => resolve(0)))) === 'Ok!'
```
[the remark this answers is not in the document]
nprimo: I replaced all the `_ => ...` functions with `() => ...` - I don't know if there is a convention, but I they both work

subjects/devops/kept-promise/README.md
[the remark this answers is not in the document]
eslopfer: We agreed to remove it since it's not needed.

subjects/devops/kept-promise/README.md
```diff
@@ -0,0 +1,56 @@
+## kept-promise
+
+### Instructions
+
+Things in your code do not always happen as fast as you would like.
+
+Sometimes it is important to wait for a function to finish before keep going on. Other times you don't want your code to be blocking while waiting.
+
+Create an asynchronous function `processInfo` in a file `kept-promise.js`. The function `processInfo` will take an asynchronous function as an input, and it will print the message `Ok!` if the return of the input function is zero or a multiple of two, `Error!` otherwise.
```
eslopfer, line 9: It would be nice to have a hint explaining what is an async function, and provide an example, in a similar way as in most exercises in the scripting piscine. For instance:
https://github.com/01-edu/public/tree/master/subjects/devops/burial

subjects/devops/kept-promise/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+The following function can be used as an example of input function for `processInfo`:
+
+```js
+const getImportantInfo = async _ =>
```
eslopfer, line 14: I know that it works, but that doesn't mean is correct, unless you need for some reason to declare a parameter that is not going to be used, it would be better to use `()`

subjects/devops/kept-promise/README.md
eslopfer: What I mean is a verbose description of what is a promise, and what parameters it takes, as in the exercises in the scripting piscine, for instance:
https://github.com/01-edu/public/tree/master/subjects/devops/burial

subjects/devops/kept-promise/README.md
eslopfer: It would also be nice to explain what does asynchronous mean in this context.

subjects/devops/kept-promise/README.md
```diff
@@ -0,0 +1,56 @@
...
+$  node main.js
+Error!
+$  node main.js
+Ok!
+$
+```
+
+### Hints
+
+- It is possible to wait for an asynchronous function with the keyword `await`. Alternatively, it can be clearer to use the method `.then`.
```
eslopfer, line 49: Here, I miss an example for instance something like this, slightly modified to resolve instead of rejecting:
```
async function f() {
  const thenable = {
    then(resolve, reject) {
      reject(new Error("rejected!"));
    },
  };
  await thenable; // Throws Error: rejected!
}

f();
```

eslopfer: It would be nice to explain what is then() as well, and an example of how it works.

subjects/devops/kept-promise/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Notions
+
+- [Introducing asynchronous JavaScript](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous/Introducing)
+- [await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await)
```
eslopfer, line 54: Since then is being mentioned, it would be nice to have a link, for instance https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then

eslopfer (changes requested):
I added some comments being more specific with what in my opinion would be better to be more verbose about, as requested :slightly_smiling_face:

subjects/devops/kept-promise/README.md
nprimo: I feel there are many examples of asynchronous processes now with the `Promise`, `await` and `then` examples. As a general concept, I feel the link to the general documentation is the best to explain this complex concept - if you have a nice, brief way to explain this to complement the official documentation, please, write it here :)

eslopfer: What would you think about this:
" Asynchronous in the context of server-side JavaScript refers to the ability of code to run non-blockingly, meaning for instance that the server can handle multiple requests simultaneously without waiting for a long-running task to complete, avoiding to block the  server responsiveness. It is essential to achieve scalable and functional back-end applications in Node.js."

subjects/devops/kept-promise/README.md
```diff
@@ -0,0 +1,72 @@
...
+Error!
+$  node main.js
+Ok!
+$
+```
+
+### Hints
+
+- A `Promise` is a special JS object that represent the result (succes or failure) of an asynchronous operation. This special object is usually used to "wrap" asynchronous operations.
+  `Promise`s can have three different states: _pending_ - the asynchronous operation has not finished yet, _fulfilled_/_resolved_ - the asynchronous operation has finished successfully - or _rejected_ - the asynchronous operation has finished, but something went wrong. When defined from scratch, it is possible to define a `resolve` and `reject` callback function for a new `Promise` that will define the results of the success or failure of asynchronous operation happening inside the `Promise`
```
eslopfer, line 48: There is a missing `.`

eslopfer (approved):
LGTM! :rocket: 

No need for extra round of reviews, once you add the explanation of what is asynchronous you can resolve the conversations and merge :+1:

PR #1864, opened 2023-03-14
DEV-4539-piscine-js-issue-with-fifty-shades-of-cold-exercice

nprimo:
- change querySelector for first test

### DIFF
--- a/dom/fifty-shades-of-cold_test.js
+++ b/dom/fifty-shades-of-cold_test.js
@@ -13,10 +13,13 @@ export const tests = []
 tests.push(async ({ page, eq }) => {
   // check that the css is properly generated
 
-  const style = await page.$$eval('style', nodes => nodes[1].innerHTML)
-  const classes = style
-    .split('}')
-    .map(s => s.replace(/(\.|{|:|;|\s+)/g, ''))
+  const style = await page.$$eval('head style', nodes =>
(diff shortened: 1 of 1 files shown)


dom/fifty-shades-of-cold_test.js
```diff
@@ -13,10 +13,13 @@ export const tests = []
...
-  const style = await page.$$eval('style', nodes => nodes[1].innerHTML)
-  const classes = style
-    .split('}')
-    .map(s => s.replace(/(\.|{|:|;|\s+)/g, ''))
+  const style = await page.$$eval('head style', nodes =>
+    nodes.map(n => n.innerText),
+  )
+
+  const classes = style[0]
+    .split('\n')
```
mikysett, line 21: This line will make valid solutions fail the test since start a new line for every new element is a best practice but has no semantic value in CSS.

It could be replaced by `.split('}')`

mikysett (changes requested):
A small modification in order to ensure valid solutions do not fail tests.

It must be said there is room for improvement since many invalid solutions may also pass the tests.
This is due to excessive usage of regex and to splitting technique to separate the different classes. Those edge cases require a more profound refactor and can be done at a later point in time.

mikysett (approved):
LGTM

PR #1873, opened 2023-03-17
fix return type of function

nik-don:
### Why?

This corrects a typing error, as 64 is not a valid type in Rust.


nprimo (approved):
@nik-don, thank you for the input. For future PRs, please, open separate PRs if they address separate exercises/projects or issues to be easier to address them.

Not a big issue since in this case we're talking about typos :)

PR #1900, opened 2023-03-27
DEV-4740: auto-jobs wrong image link

MSilva95:
remove the assets link to the image and add the image the correct way in the readme


mikysett (approved):
While the PR solves the issue it is supposed to solve I would strongly suggest to replace the image with a plain table in markdown since it will be more accessible, less prone to issues and nicer.

I rewrote it in case you want to go this way:
```
|Field|Description|Allowed Value|
|---:|---|---:|
MIN|Minute field|0 to 59|
HOUR|Hour field|0 to 23|
DOM|Day of month|1-31|
MON|Month field|1-12|
DOW|Day Of Week|0-6|
CMD|Command|Any command to be executed.|
```

PR #1904, opened 2023-03-27
DEV-4530-piscine-scripting-quest-4-division

mikysett:
Added `bc` in Dockerfile since the dependency is required for `division` exercise.
It needs to stay on its own line since it needs `curl` to be installed.


MSilva95 (approved):
LGTM!

PR #1906, opened 2023-03-28
DEV-4730-DEV-4529-fix-permission-denied-devops-tests

mikysett:
give execution permission to student files to prevent permission denied on execution

This will prevent the error from appearing again in any exercise of bash part in devops branch


MSilva95 (approved):
LGTM!

PR #1907, opened 2023-03-28
DEV-4529-piscine-scripting-quest-4-greatest-of-all

mikysett:
function is now more modular and got called once for every test
The main problem is solved by the [PR-1906](https://github.com/01-edu/public/pull/1906)

### DIFF
--- a/sh/tests/greatest-of-all_test.sh
+++ b/sh/tests/greatest-of-all_test.sh
@@ -5,25 +5,23 @@ IFS='
 script_dirS=$(cd -P "$(dirname "$BASH_SOURCE")" &>/dev/null && pwd)
 
 challenge() {
-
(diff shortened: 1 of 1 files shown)


sh/tests/greatest-of-all_test.sh
```diff
@@ -5,25 +5,23 @@ IFS='
...
+    expected=$(echo -e $1 | bash "./solutions/greatest-of-all.sh")
     diff <(echo "$submitted") <(echo "$expected")
+    if [ $? != 0 ]
+	then
+		exit 1
+	fi
 }
 
-challenge
+challenge "0\n3\n2\n5\n7\n1\n4\n9\n8\n6\n10"
```
MSilva95, line 17: It has one 11 numbers here, the greatest was supposed to be the '10' but returns "9" in the test because it has one extra number

MSilva95 (changes requested):
just a minor change, the test works the correct way, but one of the cases has one extra number.

MSilva95 (approved):
LGTM!

PR #1908, opened 2023-03-28
DEV-4743-guess-it-2-fix-wrong-info-in-the-subject

mikysett:
fix inconsistency between audit and subject about datasets used to test the project


davhojt (approved):
LGTM.

PR #1914, opened 2023-03-29
DEV-4753-dev-ops-scripting-piscine-exercises-failing-when-expected-file-doesnt-match-exercise-name

mikysett:
tests were failing when the exercise name didn't match the expected file

### DIFF
--- a/sh/tests/entrypoint.sh
+++ b/sh/tests/entrypoint.sh
@@ -6,7 +6,9 @@ cp -r /app .
 cp -a student app
 cd app
 
-chmod +x "./student/${EXERCISE}.sh"
+if test -f "./student/${EXERCISE}.sh"; then
+	chmod +x "./student/${EXERCISE}.sh"
+fi
 
 if ! test -f "${EXERCISE}_test.sh"; then
 	echo "No test file found for the exercise : $EXERCISE"


sh/tests/entrypoint.sh
```diff
@@ -6,7 +6,9 @@ cp -r /app .
 cp -a student app
 cd app
 
-chmod +x "./student/${EXERCISE}.sh"
+if test -f "./student/${EXERCISE}.sh"; then
+	chmod +x "./student/${EXERCISE}.sh"
+fi
```
HarryVasanth, line 11: Would this be better?

```bash
find /student -type f -name "*.sh" -exec chmod +x {} \;
```

mikysett: Thanks a lot for the suggestion, actually I initially tough about this approach, but I'm not sure we want all files to be executable (my fear is that some exercises are indeed about `chmod` and we may have random failures by forcing a specific state on all files in there)

HarryVasanth: Ah, ok then it should be good!

HarryVasanth:
@mikysett  This will resolve most of the issues, except in the cases where non-executable bash scripts are run with `./script.sh` instead of `bash script.sh` command.

HarryVasanth (approved):

PR #1915, opened 2023-03-29
DEV-4754-gatecrashers-node-js-quest-wrong-info-about-previous-exercise

mikysett:
The exercise refers to `late-guests` which was the old name of `uninvited` exercise


davhojt (approved):
LGTM.

PR #1918, opened 2023-03-30
DEV-4757 resolve the inconsistency between the audit and the subject of push swap

zanninso:
must fix the inconsistency audit and the subject as specified  in the issue conversation in GitHub between   and one of the learners there is an inconsistency between the audit and the subject of push-swap project  in the Go brach, which the audit deal with the checker as executable and the subject deal with is a program.


MSilva95 (approved):
LGTM!

PR #1936, opened 2023-04-05
Add simplified matrix explanation

nik-don:
This commit provides a simplified, non-technical description of the matrix data structure and its basic operations for Rust implementation. The explanation is tailored for those who might be new to programming or unfamiliar with matrix concepts.


nprimo:
Hi @nik-don, thank you for your feedback and input. For the moment, we'll keep the current text: if some of the words or concepts presented are not clear to someone, they're encouraged to search for those to get familiar with those terminologies.

PR #1958, opened 2023-04-13
DEV-4788-HelloWorld-Java-Exercise

zanninso:
Adding all subjects of the first quest of the java piscine


subjects/java-piscine/Quest1/01-02-int-operation/README.md
```diff
@@ -0,0 +1,76 @@
...
+Multiply
+18
+132
+Divide
+4
+3
+$ 
+```
+
+# Notions
```
MSilva95, line 74: wrong hashes should be 3

subjects/java-piscine/Quest1/01-02-int-operation/README.md
```diff
@@ -0,0 +1,76 @@
...
+        System.out.println(IntOperation.multiplyTwoIntegers(12, 11));
+        
+        System.out.println("Divide");
+        System.out.println(IntOperation.divideTwoIntegers(8, 2));
+        System.out.println(IntOperation.divideTwoIntegers(13, 4));
+    }
+}
+```
+
+and its output :
```
MSilva95, line 55: You should install prettier so that you can format the readmes and avoid extra spaces like in line:
32,39, 43, 47, 58,71, 75

and it helps with the format of the radme example:
image
How it looks after 
image

subjects/java-piscine/Quest1/01-02-int-operation/README.md
```diff
@@ -0,0 +1,76 @@
+## IntOperations
```
MSilva95, line 1: if you use `int-operations` in the folder naming you should use `int-operations` as title instead of `IntOperations`
pick one and be consistent for every exercise, in Java, if I'm not mistaken the naming convention that is mostly used is the camelCase so i would go with that...

subjects/java-piscine/Quest1/01-03-float-operation/README.md
```diff
@@ -0,0 +1,58 @@
...
+Add
+3.5
+83.46347
+Divide
+3.5
+0.8115942
+$
+```
+
+# Notions
```
MSilva95, line 55: wrong hashes should be 3

subjects/java-piscine/Quest1/01-04-is-even/README.md
```diff
@@ -0,0 +1,49 @@
...
+```shell
+$ javac *.java -d build
+$ java -cp build ExerciseRunner
+true
+true
+false
+$
+```
+
+# Notions
```
MSilva95, line 45: wrong hashes should be 3

subjects/java-piscine/Quest1/01-11-is-palindrome/README.md
```diff
@@ -0,0 +1,47 @@
...
+true
+false
+$ 
+```
+
+### Notions
+[String](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/String.html)  
+[Conditional statement](https://docs.oracle.com/javase/tutorial/java/nutsandbolts/if.html)  
+[Loop statement](https://docs.oracle.com/javase/tutorial/java/nutsandbolts/for.html)
+
```
MSilva95, line 43: avoid this extra lines, with `prettier` it solves this

MSilva95 (changes requested):
For most of it you got it right. But there is some small mods you need to do:

- The main folder should be named `java-piscine` and then the exercises should be inside, no need to add the folder `Quest1, quest2 ...` all the exercises readmes will be at the same level.
- The names of the folders should not be `01-01-helloworld` only `helloworld` the order of it is given in the refs so find a way to save that order in a doc for you to use it latter
- The folder name should be camel case and it must be the same as the title in the readme
- Pay attention to the hashes so that you keep the same format for all the readmes

MSilva95 (changes requested):
Apart from the inline comments LGTM. After fixing the inline comments you can merge it.

MSilva95 (approved):
LGTM!

PR #1963, opened 2023-04-17
DEV-4890 - add prettier check for new ".md" files

nprimo:


mikysett:
After checking together we found out this GA was problematic under some aspects.
In general its hard to test it locally, when trying with `Act` we managed to make a more simple version that seems to work as expected:

```
name: 'Prettier check'
on:
  workflow_dispatch:
  pull_request:
    branches:
      - master
jobs:
  build:
    runs-on: ubuntu-latest
    name: Run prettier on changed files
    steps:
      - uses: actions/checkout@v3
        with:
          fetch-depth: 2  # OR "2" -> To retrieve the preceding commit.

      - name: Get changes
        id: changed-md
        run: echo "changed_files=$(git diff --name-only -r HEAD^1 HEAD | xargs)" >> $GITHUB_OUTPUT

      - name: Run step if any *.md file(s) changed 
        # if: steps.changed-md.outputs.any_changed == 'true'
        run: |
          npm i -gs prettier
          echo "====> FILES CHANGED: " ${{ steps.changed-md.outputs.changed_files }}
          for file in ${{ steps.changed-md.outputs.changed_files }}; do
              echo "$file was changed"
              npx prettier -c $file
          done
```

This solution is very incomplete, missing a grep for `.md` and a proper `if` statement to trigger prettier.

HarryVasanth:
In this (new) iteration, we are using the same prettier action that we use on repositories across our org. It will check for changed files within the PR to see if any 💄 Prettifying needed.

#### File extensions that are covered:
- js
- md
- yml
- json
- sh

mikysett (approved):
LGTM (tested together locally with `act`)

PR #1979, opened 2023-04-20
js-piscine:pimp-my-style - fix mistyping Er're -> We're

maggieeagle:
Small fix to pimp-my-style, fixed mistyping to make task text more grammar correct


nprimo:
Thank you for your input @maggieeagle! This typo has been already fixed :)

PR #2021, opened 2023-05-16
DEV-5013 to-git-or-not-to-git remake

MSilva95:
### Why?

Due to some changes on the permissions for the students, this exercise was no longer  possible to complete

### Solution Overview

 it was remade and the order that it appears on the quest was changed.

### Implementation Details


nprimo (changes requested):
I would add a check on the `_test.sh` file to avoid solutions that are `echo $expeceted_solution`

PR #2033, opened 2023-05-24
DEV-5029-add-readme-requirement-for-crud-master

mikysett:


zamazzal:
Perfect, but don't merge it right now until we get approval from Cyril.

PR #2046, opened 2023-05-31
DEV-3241-DEV-3242-corewar-subject-and-audit

mikysett:
First complete draft of subject and audit.

While content should be exhaustive and complete there may be still a lot of errors to check and room for improvement.

Many internal links are not supposed to work yet or are simply not written down still.


subjects/corewar/README.md
```diff
@@ -0,0 +1,366 @@
...
+- The Virtual Machine and the Assembler can be done in any compiled language.
+- You should only use the standard library of the language you choose.
+- In case you use any external library it must be largely justified and explained (avoid it at any cost).
+- You must not use external libraries to do the parsing part of the Assembler for any reason.
+- The programs should never crash for any reason.
+- The programs should never leak memory for any reason.
+
+### Game dynamics
+
+At its very core the game consists of having two or more programs (written by the contestants) that will be executed by the Virtual Machine in an arena (a sandbox memory space).
```
nprimo, line 22: I would already use a term like `players` (or whatever we think it works best) instead of program

subjects/corewar/README.md
```diff
@@ -0,0 +1,366 @@
...
+The instructions will be of different sizes depending on the number of type of parameters passed.
+
+#### Labels
+
+Labels are also called pseudo instructions, in the sense they won't be translated into binary code.
+They are widely used to help programmers creating assembly code without needing to remember and count relative memory positions.
+
+So when looking in the binaries the labels will be replaced in order to match their relative position in bytes from the place they were used.
+
+As an example in the `ameba.s` player `zjmp %:hello` is exactly the same of `zjmp %-5` (`-5` being `ff fb` in hexadecimal notation).
```
nprimo, line 150: `ameba.s` is not introduced before this point - it might be easier to have the `ameba.s` example right after the section **### The Assembly language*** or at least a reference to jump to the example snippet

subjects/corewar/README.md
```diff
@@ -0,0 +1,366 @@
...
+> Notice that some smart players may trick another player to start making `live` statements for them, this virtually means a player can still make `live` statements after all its processes were killed.
+
+### The Assembly language
+
+In the assembly file there must be a `.name "[Name of the player]"` field and a `.description "[Description of the player]"` before any instruction.
+
+Every instruction has to be on a single line and may be preceded by a label declaration.
+The instruction name is separated from the optional label and the parameters by one or more spaces.
+
+The parameters are separated from each other by `,`.
```
nprimo, line 139: I would add something like: 
```
Instructions parameters can be *Registers*, *Direct Arguments* or *Indirect Arguments*.
``` 
The information is indirectly passed later on, but I feel this would reduce the "learning fatigue"

subjects/corewar/README.md
```diff
@@ -0,0 +1,366 @@
...
+> Notice that labels can be declared at the start of an instruction or on their own line.
+
+#### Parameters types
+
+- Register parameter
+
+Each process has its own registers (from 1 to `REG_NUMBER`).
+Each of those registers can hold 32 bits, but the register number itself is written in only 8 bits (which are enough to go from 1 to 16).
+
+They are formatted as `r1`, `r2`, `...`, `r16`.
```
nprimo, line 163: It's quite straightforward, but to be sure I would add that when passed as an argument, the value will be the one stored in the specified register. Additionally I would add which values are stored when the VM is initialized

subjects/corewar/README.md
```diff
@@ -0,0 +1,366 @@
...
+| zjmp  | 1         | 9      | 20        | false     | true    | Direct                                                                     |
+| ldi   | 3         | 10     | 25        | true      | true    | [Register, Indirect, Direct]<br> [Register, Direct]<br> Register           |
+| sti   | 3         | 11     | 25        | true      | true    | Register<br> [Register, Indirect, Direct]<br> [Register, Direct]           |
+| fork  | 1         | 12     | 800       | false     | true    | Direct                                                                     |
+| lld   | 2         | 13     | 10        | true      | false   | [Indirect, Direct]<br> Register                                            |
+| lldi  | 3         | 14     | 50        | true      | true    | [Register, Indirect, Direct]<br> [Register, Direct]                        |
+| lfork | 1         | 15     | 1000      | false     | true    | Direct                                                                     |
+| nop   | 1         | 16     | 2         | true      | false   | Register                                                                   |
+
+
```
nprimo, line 198: General minor comments for all the below points: whenever there is a "bracket statement" to close the brief description I would either (1) always use capital letters for the first word inside the parenthesis or (2) move the `.` at the end of the brackets and keep lower case for the first word in the parenthesis

subjects/corewar/README.md
```diff
@@ -0,0 +1,366 @@
...
+In the assembly file there must be a `.name "[Name of the player]"` field and a `.description "[Description of the player]"` before any instruction.
+
+Every instruction has to be on a single line and may be preceded by a label declaration.
+The instruction name is separated from the optional label and the parameters by one or more spaces.
+
+The parameters are separated from each other by `,`.
+
+The instructions will be of different sizes depending on the number of type of parameters passed.
+
+#### Labels
```
nprimo, line 143: I would start with the instruction table maybe and place the Label section toward the end of this "higher" level section about the assembly language

subjects/corewar/README.md
```diff
@@ -0,0 +1,366 @@
...
+
+An indirect parameter will give you the relative position of the address where to look for the value, think of it as an ancestor of raw pointers.
+
+In this case `10` will say to the VM to look ten bytes forward the current instruction and take the value from there.
+
+> Don't be scared to use the provided Asm and VM to play with those different parameters in order to be sure you understood properly their behavior.
+
+#### Instruction Set
+
+| Name  | Nb Params | Opcode | Nb Cycles | Has Pcode | Has Idx | Params                                                                     |
```
nprimo, line 179: It might be a bit repetitive, but I would add a short description for each column: some information is already present in the below paragraph while some are missing, for example, `Nb Params` or `Params`. I am convinced is it possible to understand what each of the column means, but it would help reducing doubts

subjects/corewar/README.md
```diff
@@ -0,0 +1,366 @@
...
+
+#### The Carry flag
+
+In all processors you will find a carry which is a flag that notify some edge case states to allow special operations, ensure security and serve many others functionalities.
+
+The Carry of the processes in the VM is in that sense a simplified version of this concept.
+
+The following commands will modify the carry of the process executing them: `ld`, `add`, `sub`, `and`, `or`, `xor`.
+
+Those commands will set the carry to `true` if the result of their respective operations is `0`, and set it back to `false` otherwise.
```
nprimo, line 250: This is not clear to me: it is never introduced the concept of `the result of their respective operation` before this point - how is defined the result of an instruction? As far as I know, the state should depend on the first argument passed to this instruction (but I might be completely wrong)

subjects/corewar/README.md
```diff
@@ -0,0 +1,366 @@
...
+- Instructions on how to execute the references.
+- Some players to use as tests.
+
+> The provided VM has an extra flag `-v` which you can use to print the state of the VM at every cycle, this should greatly help you during development and debugging.
+
+### Some advices for the road
+
+While the functioning of the VM and the Assembler may seems quite strange and obscure, it is a perfect scenario to reproduce a CPU and understand what a computer really does at its core.
+
+In this project you are indeed creating a Turing Complete machine largely inspired by Von Neumann architecture (which is still how today's processors works).
```
nprimo, line 356: Maybe add some links to wikipedia pages:
- https://en.wikipedia.org/wiki/Turing_machine
- https://en.wikipedia.org/wiki/Von_Neumann_architecture

subjects/corewar/audit/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+##### Run the learner VM without any argument.
+
+###### Does the VM prints a little introduction on how to use it?
+
+##### Run the VM with a path to a file that doesn't exist.
+
+###### Does the VM returns a meaningful error and exit?
+
+###### Does the VM returns a meaningful error and exit?
```
nprimo, line 51: This line is repeated

nprimo (changes requested):
I left some comments inline - overall it looks very good!

subjects/corewar/audit/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+##### Run the VM with four valid players and the flag `-d 10`.
+
+###### Does the VM introduce each player by writing their name and description before to start the game?
+
+###### Does the VM exit at cycle 10 and dump the memory printing 32 bytes per row?
+
+##### Run the VM using four players with the name starting with `strong_*` and the flag `-d 40000`.
```
nprimo, line 59: Minor comment: if we provide the student with "strong" players then it would be easier to tweak one of these players to win vs `ameba.s` -

subjects/corewar/README.md
mikysett: I've swapped `Labels` and `Parameters type`. Now the definition of the parameters is just after this part

subjects/corewar/README.md
mikysett: I'm not sure I understood so please correct me if I misread some part.

- The register can only pass the value that it contains... if it would pass it's register number instead it would just be a direct value (`%4` instead of `r4` for example).
On one side I understand at first glance it may not be obvious, still, adding more words to explain that may add confusion and we would be explaining why you can't do something that doesn't makes sense, which I'd rater prefer students to figure out by themselves.

- For the initialization of the register, I've added it ;)

subjects/corewar/README.md
mikysett: done to uppercase

subjects/corewar/README.md
mikysett: done: moved labels after the IS

subjects/corewar/README.md
mikysett: done

subjects/corewar/audit/README.md
mikysett: it is true, still to create a strong player is not the scope of this project, even if they use this strategy it is fine: they will read and understand a player source code, which is all we want from them at that moment

subjects/corewar/README.md
mikysett: it is true this part is not linear, still `ameba` will be present in the playground and it is also fully translated and explained later in the subject.

To delay this part means having less info when the labels are explained or to explain labels at a very later point (where they would have shown up already multiple times, reproducing the same issue at a bigger scale).

Still there may be other ways to go, if you feel we can rephrase or move just part of those paragraphs let me know, I don't have strong opinions on that one

subjects/corewar/README.md
```diff
@@ -0,0 +1,370 @@
...
+- The Virtual Machine and the Assembler can be done in any compiled language.
+- You should only use the standard library of the language you choose.
+- In case you use any external library it must be largely justified and explained (avoid it at any cost).
+- You must not use external libraries to do the parsing part of the Assembler for any reason.
+- The programs should never crash for any reason.
+- The programs should never leak memory for any reason.
+
+### Game dynamics
+
+At its very core the game consists of having two or more programs (written by the contestants) that will be executed by the Virtual Machine in an arena (a sandbox memory space).
```
nprimo, line 22: I would specify here or later on that the VM would only take up to 4 players at the same time

mikysett: done on the VM part when defining the CL arguments

subjects/corewar/README.md
mikysett: normally they already are hyperlinks

nprimo (approved):
LGTM! Well done, it looks very clear and quite short considering the amount of information that is required for the project :muscle:

PR #2057, opened 2023-06-19
 DEV-5079 Improve Public tests CI/CD

HarryVasanth:
### Why?

Currently, CI/CD actions are working on both PRs and Master branch at the same time, so if a PR fails, the master is “broken” 

### Solution Overview

Separate CI/CD actions to distinguish the differences. 
Applies to `test-sh`, `test-js` and `test-dom`

### Implementation Details

3 Different CI/CD actions:

- ga-image-sanitize
- ga-image-build-master
- ga-image-sanitize


MSilva95 (approved):
LGTM!

PR #2064, opened 2023-06-22
CON-1634 update corewar subject

nprimo:
Make subject consistent with new VM version

### DIFF
--- a/subjects/corewar/README.md
+++ b/subjects/corewar/README.md
@@ -44,7 +44,8 @@ The players will be able to create new processes using specific instructions, it
 
 There is a special instruction, called `live` that notifies the VM one player is still healthy and doing well.
 
-Every once in a while the VM will check if the players executed such instruction and, if they didn't, all the processes created by this player will be instantly killed.
+Every once in a while the VM will check which processes executed such instruction and kill all the processes that have not.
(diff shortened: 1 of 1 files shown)


subjects/corewar/README.md
```diff
@@ -44,7 +44,8 @@ The players will be able to create new processes using specific instructions, it
 
 There is a special instruction, called `live` that notifies the VM one player is still healthy and doing well.
 
-Every once in a while the VM will check if the players executed such instruction and, if they didn't, all the processes created by this player will be instantly killed.
+Every once in a while the VM will check which processes executed such instruction and kill all the processes that have not.
+After this check, the VM will consider that every remaining process is dead until it successfully executes a `live` instruction.
```
mikysett, line 48: This line may be confusing... Considering a process to be "dead" may means this process should not perform any action, maybe we can rephrase it to better explain the expected behavior.

mikysett (changes requested):
The section `#### Stop player execution` at line `123` is not valid anymore and should be modified according the new rules.

Maybe it could be nice to change the title also into `#### Stop process execution`

subjects/corewar/README.md
nprimo: I rephrased it

PR #2069, opened 2023-06-30
devops cloud : cloud-design Update README.md

lotoussa:
remove duplicated paragraph


nprimo:
@lotoussa it has already been removed :)

PR #2084, opened 2023-07-12
fix(java-piscine): change the name of the solution class form Sort to…

zanninso:
change the name of the solution class form Sort to SortList


akhossanX (approved):
Seems good

PR #2086, opened 2023-07-13
fix: edit signature of mapByFirstLetter to make it consistent with th…

zanninso:
edit signature of mapByFirstLetter to make it consistent with the tests


akhossanX (approved):
Good

PR #2099, opened 2023-07-25
CON-1810-MARKDOWN-drawing-raid-for-Java-piscine

Hamzaelkhatri:
docs(Jart):add the audit of jart raid

### Why?

### Solution Overview

### Implementation Details


subjects/java/raids/Jart/README.md
zanninso, line 1: the interface Displayable its not used here at all you can remove it while the image.save used to result the image or use display from Displayable interface to produce the image and require for the student to produce the image by opening a file and writing the pixels there you can use the easiest image format for not make it hard

subjects/java/raids/Jart/audit/README.md
zanninso, line 1: ask the students about the interfaces and if it's well used and implemented 
you can ask also if the student use a default implementation of in the interfaces it self

zanninso (changes requested):
please resolve the added conversations

subjects/java/raids/Jart/README.md
Hamzaelkhatri: The Displayable interface is implemented in the Image class. the method `save` for example is to get it from the Displayable interface

zanninso (changes requested):

subjects/java/raids/Jart/README.md
Hamzaelkhatri: from the example, the student should understand that the `Image` class implements `Displayable`. and Displayable contains two functions `display` and  `save`

```java
interface Displayable {
    void display(int x, int y, Color color);

    void save(String string);
}
```
display function used inside `Drawable` as a parameter of the `draw` function 

```java
interface Drawable {
    void draw(Displayable displayable);
   //...
}
```

> I add the `save`  function in the subject in the part of  `Displayable` interface to make it more clear.

zanninso: there is no details of the interfaces methods signature in the subject and this make it harder even for me to understand, and in this way we keep a gap where students can do a lot of mistakes you have to specify the signatures of the interfaces methodes, like in the example same the object of the image into a variable of type displayable this way can give a hint to the students about the signatures and parametres types.
to recap here is what you have to change:
- add the signature of the display method that will take the corrdinates`x, y` and a color
- change the image variable type to `Displayable `

subjects/java/raids/Jart/audit/README.md
Hamzaelkhatri: Done 👍🏻

zanninso (approved):
LGTM

PR #2138, opened 2023-08-22
docs(SortArgs) - Fix bug in ExerciseRunner

maximihajlov:
Restores the default `System.out` to fix result printing

`System.out` was changed to a custom `printStream`, so 

```Java
System.out.println(output.equals("1 2 3 4\n"));
```

printed `true` or `false` to this `printStream` and not the default `System.out`. This caused console output to be empty.


zanninso (approved):
LGTM

PR #2141, opened 2023-08-23
CON-1973-Update-Subject-and-Audit-of-atm-management

mikysett:
Fix an error defining wrong pourcentages (examples: 0.07% instead of 7%)


zainabdnaya (approved):
Looking good to me👍.

PR #2157, opened 2023-08-29
CON-1994-Update-subject-for-card_deck-rand-crate-dependency

mikysett:
CAUTION: The subject and the tests (`rust-tests`) should be merged at the same time!


mikysett:
This PR is closed because `rand` crate is used more widely and a new PR will fix and all of those dependencies

PR #2177, opened 2023-09-08
Con 1813 markdown chaikin raid for java piscine

zanninso:


subjects/java/raids/Jaikin/README.md
nprimo, line 1: Is there any reason why the project is in `raids` directory? I would keep all the `java` content in the `java` directory without adding extra directories inside. The fact that this is a `raid` would be specified later on in the `ref`

subjects/java/raids/Jaikin/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+Optionally, you may implement the following bonus features:
+
+1. Allow the user to clear the screen, enabling them to select new control points.
+
+2. Implement real-time dragging of the control points, so the user can adjust their position interactively.
+
+> Note: You are free to choose any library for handling windows, rendering, keyboard, and mouse events.
+
+For more information on Chaikin's algorithm, refer to [this resource](http://graphics.cs.ucdavis.edu/education/CAGDNotes/Chaikins-Algorithm/Chaikins-Algorithm.html). You can also view an example of how the application should work in this [video](https://youtu.be/PbB2eKnA2QI). Remember, the usage of Chaikin's algorithm is mandatory for this implementation.
```
nprimo, line 37: I cannot access the link `http://graphics.cs.ucdavis.edu/education/CAGDNotes/Chaikins-Algorithm/Chaikins-Algorithm.html` - I get a 403. 

After a quick google search, I found this link that gives some information on how to implement the algorithm. https://www.cs.unc.edu/~dm/UNC/COMP258/LECTURES/Chaikins-Algorithm.pdf

subjects/java/raids/Jaikin/README.md
nprimo, line 1: The algorithm is called `chaikin` - why is the name of the project `jaikin`?

subjects/java/raids/Jaikin/README.md
```diff
@@ -0,0 +1,37 @@
+## Implementation of Jaikin's Algorithm with Step-by-Step Animation
+
+### Objective
```
nprimo, line 3: I feel that it could be nice to add that the goal of the project is to implement a program that `... implement Chaikin's algorithm...`

I understood this only after viewing the video or reading the `audit`

subjects/java/raids/Jaikin/audit/README.md
nprimo, line 1: This audit is not correctly formatted - please re-format according to 01 standards. 
For example [`go-reloaded` audit](https://github.com/01-edu/public/blob/master/subjects/go-reloaded/audit/README.md?plain=1) and you can use https://preview.01-edu.org/preview to ensure that the format is correct.

All the points "verify" should have a `yes/no` type of quesiton.

nprimo (changes requested):
I left some comment inline + the 2 files needs to be formatted using `prettier`

subjects/java/raids/Jaikin/README.md
zanninso: this is how the java folder structured and i see that this way is better for having a clear separation

zanninso: because it's the same raid that we have in rust so we cannot use the same name, I replaced the ch with j to show that is a java project and to have some kind of consistency between this raid and other java raids

subjects/java/raids/Jaikin/README.md
zanninso: did you read the objective of the project ? it's clearly mentioned there !!

subjects/java/raids/Jaikin/README.md
```diff
@@ -0,0 +1,37 @@
+## Implementation of Jaikin's Algorithm with Step-by-Step Animation
```
nprimo, line 1: I would rename this as `Jaikin` or change the sentence by saying 
`Implementation of Chaikin's .....` 

based on what you shared [here](https://github.com/01-edu/public/pull/2177#discussion_r1513044407)

subjects/java/raids/Jaikin/README.md
nprimo: In the `Objective` it is mentioned that the raid would be about implementing an algorithm, but not how it would be used: it could be a web app, a function, a CLI tool, ... 

It is clear after checking the audit and video that what it is expected is a program, but adding the information here as well could be nice

zanninso: This is what we have in the project objective ```The goal of this project is to implement Chaikin's algorithm and create a step-by-step animation of the process using a canvas``` we clearly mentioned that they have to impliment the chaikin's algorithm and they have to use canvas and java to do that so it's clear that it would be a program and not a web page or video or anything else

nprimo: This is the first result I found, looking for `canvas programming`: https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial

Without looking all the subject and the video provided, if I had to do this project, I could easily start looking on how to develop a web app and not a program (as it seems to be the case).

nprimo (approved):
LGTM!

You can resolve the remaining conversation - I have added more context to justify my previous comment: I am sure students can understand the required output type by reading all the information provided.

subjects/java/raids/Jaikin/README.md
zanninso: I see that you forget that this project will build with java so if they have to search they have to search about canvas in java

PR #2189, opened 2023-09-13
docs: add subject and audit files

zanninso:
adding the subject and the audit file for the mr-jenk project.


akhossanX (approved):
Quite good

PR #2213, opened 2023-09-21
CON-2148-luhn_algorithm-subject

mikysett:


subjects/luhn_algorithm/README.md
```diff
@@ -0,0 +1,62 @@
+## luhn_algorithm
+
+### Instructions
+
+Create a function which checks if a number is valid per the Luhn formula.
+
+The function will receive a string and return a boolean.
```
MSilva95, line 7: This section is a bit confusing maybe we can separate it like this:

```
The function will receive a string as its input and return a boolean value and you'll have to take into account the following:

**Invalid Inputs:** It categorizes an input as invalid if it is an empty string or a number with only one digit.

**Handling Spaces:** While spaces within the input are allowed, they are automatically removed during the calculation process, ensuring they do not impact the result.
```

subjects/luhn_algorithm/README.md
```diff
@@ -0,0 +1,62 @@
...
+- We multiply those digits by 2
+- If the result is more than 9 we subtract 9 from it
+- We sum all the digits
+- If sum is evenly divisible by 10 then this number is valid
+
+So we will get:
+
+- `4539 3195 0343 6467`
+- `4_3_ 3_9_ 0_4_ 6_6_`: numbers to modify
+- `8_6_ 6_9_ 0_8_ 3_3_`: modified numbers
```
MSilva95, line 26: this instruction is missing:
```
Handle Results Over 9: If the result of any multiplication is greater than 9, subtract 9 from it.
```

MSilva95 (changes requested):

subjects/luhn_algorithm/README.md
mikysett: done

subjects/luhn_algorithm/README.md
mikysett: done

MSilva95 (approved):
LGTM!

PR #2221, opened 2023-09-25
docs(rust piscine): Add invert_sentence optional exercise subject

akhossanX:


subjects/invert_sentence/README.md
```diff
@@ -0,0 +1,36 @@
...
+### Usage
+
+Here is a possible runner to test your function :
+
+```rust
+use invert_sentence::invert_sentence;
+
+fn main() {
+    println!("{}", invert_sentence("Rust is Awesome"));
+    println!("{}", invert_sentence("   word1     word2  "));
```
mikysett, line 23: In my original idea I was thinking about preserving the number of white spaces in between words, which should make the exercise a bit harder to solve, otherwise we should classify it as easy rather than medium.

As an example word1   word2  =>   word2   word1

mikysett (changes requested):
maybe we can make the exercise a bit more challenging

subjects/invert_sentence/README.md
```diff
@@ -30,7 +32,7 @@ And its output:
 ```console
 $ cargo run | cat -e
 Awesome is Rust$
-word2 word1$
+   word2     word1  $
```
mikysett, line 35: Here it seems like the spaces are wrong... the first one should be 2 spaces and not 3 at the start and 3 and not 2 at the end:
`  word2     word1   `

mikysett (changes requested):
a little typo to fix in the example

subjects/invert_sentence/README.md
akhossanX: indeed

mikysett (approved):
LGTM

PR #2258, opened 2023-10-18
CON-2264 Changes in piscine-go exam function exercises

MSilva95:
The main reason behind this is to address a specific issue concerning exercises in an exam that only accepts “functions”, where the expected file is not an entire program but rather an exported function, students lack a straightforward method to test their exercise.

Currently, students are required to modify the function contained in the file to make it execute as a standalone program. The proposed structure helps the student to test their work

In the new structure, the public repo will contain the exercise readme file and also a “main.go”  that will take the main necessary to test the function.

### DIFF
--- a/subjects/btreedeletenode/main.go
+++ b/subjects/btreedeletenode/main.go
@@ -1,0 +1,18 @@
+package main
+
+import (
+	"fmt"
+)
+
+func main() {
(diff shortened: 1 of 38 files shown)


LEEDASILVA (approved):
LGTM, We will need to do the same for:
- The module check-point
- All rust exercises that returns only functions on the exams

PR #2267, opened 2023-10-25
CON-2274: adding a new file for code editor to use in the rust-piscine

MSilva95:


LEEDASILVA (approved):
I see that you caught some inconsistencies on the READMEs'. Nice !

---

Checked the list of refs that we added the `codeEditor.providedFiles` and it matches the files changed here. :+1:

PR #2283, opened 2023-11-17
Update README.md

blueskiy01:


MSilva95 (commented):
According to the creator of the branch the 3 persons should not be in the program so i don't see the point in contradicting this.
Thank you for your feedback!

PR #2314, opened 2023-11-27
CON 2304 add link to dataset for first raid in AI branch

nprimo:


subjects/ai/backtesting-sp500/audit/README.md
```diff
@@ -38,7 +38,9 @@ project
 
 ###### Does the notebook contain a Histogram of average price for companies for all variables (saved the plot with the images)? This is required only for **prices.csv** data.
```
nprimo, line 39: The issue refers to this as well - it should probably be clarified in the subject that it is required to add an `histogram` with the specifics described here. Additionally, we can rephrase this, for example:

```
Does the notebook contain a Histogram that shows the average price for each company in the dataset? ...
```

PR #2353, opened 2023-12-11
CON-2331 update `push-swap` audit commands to run to be in line with subject

nprimo:
Update audit commands to be in line with the subject to avoid unexpected behaviour from different shell


MSilva95 (approved):
LGTM!

PR #2355, opened 2023-12-12
CON-2332 update subject and audit of exercise 1 of `nlp-spacy`

nprimo:
- clarify required goal of exercise.
- update audit question to avoid floating issues.


MSilva95 (approved):
LGTM!

PR #2364, opened 2023-12-26
Flags are most likely in wrong order for the audit file

ratsepmikk:
- It makes more sense for the flag -p (short for 'ports') to be before the port range
- It makes more sense for the flags -u or -t (short for 'UDP ports' and 'TCP ports') to be either freely placed where-ever or to be placed before the 'ip_address' argument, but not before the port range argument. This mistake happened only inside the audit questions file, it seems to be fine in the task description file.


nprimo (approved):
@ratsepmikk, thank you for your contribution :)

PR #2379, opened 2024-01-15
CON-2367: provide `crud-master` source code. Add bonus to use learner's own solution

nprimo:
- update audit to reflect changes in the subject

### DIFF
--- a/subjects/devops/code-keeper/README.md
+++ b/subjects/devops/code-keeper/README.md
@@ -101,9 +101,9 @@ consistent deployments of the microservices.
 
 ![code-keeper](resources/code-keeper.png)
 
-1. You will use your `crud-master` source code and `cloud-design`
-   infrastructure, to create a complete pipeline for the following
-   applications:
+1. Use the provided [`crud-master` source
+   code](https://github.com/01-edu/crud-master-py), to create a complete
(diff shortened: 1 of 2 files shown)


subjects/devops/code-keeper/README.md
```diff
@@ -220,6 +220,7 @@ you can implement anything that you feel deserves to be a bonus, for example:
 - Add `Infracost` in your infrastructure pipeline to estimate the
   infrastructure cost.
 - Use `Terragrunt` to create multiple Environments.
+- Use your own `crud-master` source code and `cloud-design` infrastructure.
```
zamazzal, line 223: using cloud-design infrastructure is mandatory , since we are not providing the solution

nprimo: Hi @zamazzal, I see, but at the moment we don't have any `terraform` / `cloud-design` solution to provide the student: is there a way we could provide a starter code that does not include any IaC? Should we provide the solution of `orchestrator` instead of `crud-master`?

nprimo: I have updated the text as we discussed in our meeting today

zamazzal (approved):
LGTM!

PR #2415, opened 2024-01-29
CON-2392 add files to the readme used in the checkpoint

MSilva95:
Adding the necessary files to the exercises used in the checkpoints

### DIFF
--- a/subjects/btreedeletenode/main.go
+++ b/subjects/btreedeletenode/main.go
@@ -2,22 +2,59 @@ package main
 
 import (
 	"fmt"
+	"piscine"
 )
 
-type TreeNode struct {
-	Left, Right, Parent *TreeNode
-	Data                string
-}
-
 func main() {
-	root := &TreeNode{Data: "4"}
+	root := &piscine.TreeNode{Data: "4"}
 	BTreeInsertData(root, "1")
 	BTreeInsertData(root, "7")
 	BTreeInsertData(root, "5")
 	node := BTreeSearchItem(root, "4")
 	fmt.Println("Before delete:")
 	BTreeApplyInorder(root, fmt.Println)
-	root = BTreeDeleteNode(root, node)
+	root = piscine.BTreeDeleteNode(root, node)
 	fmt.Println("After delete:")
 	BTreeApplyInorder(root, fmt.Println)
+}
+
+func BTreeInsertData(root *piscine.TreeNode, data string) *piscine.TreeNode {
+	if root == nil {
+		return &piscine.TreeNode{Data: data}
+	}
+
+	if data < root.Data {
+		root.Left = BTreeInsertData(root.Left, data)
+		root.Left.Parent = root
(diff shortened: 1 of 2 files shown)


subjects/btreedeletenode/main.go
```diff
@@ -4,6 +4,11 @@ import (
 	"fmt"
 )
 
+type TreeNode struct {
```
nprimo, line 7: Shouldn't we avoid the definition of the `TreeNode` type and assume the student will define it?

subjects/sortedlistmerge/main.go
```diff
@@ -0,0 +1,53 @@
+package main
+
+import (
+	"fmt"
+	"piscine"
+)
+
+type NodeL struct {
```
nprimo, line 8: In the subject and in the rest of the `main` function this structure is called `NodeI` (if I understood well)

subjects/sortedlistmerge/main.go
```diff
@@ -0,0 +1,53 @@
...
+	Data interface{}
+	Next *NodeL
+}
+
+type List struct {
```
nprimo, line 13: This type is not used in this code snippet.

nprimo (changes requested):
I would avoid adding the type definition in the `main.go`. I would keep the code as the one in the subject - or is there a reason to add the type definition?

nprimo (changes requested):
I would avoid adding the type definition in the `main.go`. I would keep the code as the one in the subject - or is there a reason to add the type definition?

subjects/btreedeletenode/main.go
MSilva95: In the piscine the have the TreeNode defined in the firsts exercises of the quest but in the exam they wont have access to the piscine so we need to give the TreeNode so that they can test the code in the code editor

MSilva95: Removed the TreeNode and added the functions they need to test with the main.go

subjects/sortedlistmerge/main.go
MSilva95: my mistake, its fixed

subjects/sortedlistmerge/main.go
MSilva95: removed

subjects/sortedlistmerge/main.go
```diff
@@ -0,0 +1,48 @@
+package main
+
+import (
+	"fmt"
+	"piscine"
+)
+
+type NodeI struct {
```
nprimo, line 8: To define their function it's better that the students define this structure in their solution. If that's not the case, every time there's a `piscine.NodeI` here should be changed to `NodeI`

nprimo (changes requested):
LGTM! :rocket:

nprimo (approved):
LGTM

PR #2421, opened 2024-01-30
CON-2396 files for checkpoints

MSilva95:
Adding imports to the main.go files used in the checkpoints


Oumaimafisaoui (approved):
Approved.

PR #2429, opened 2024-01-31
CON-2401 clarify when startup message should be print in `friend-support`

nprimo:


Pomog (approved):
Ensure correct behavior of the friend-support server tests

- Ensure the server listens on port 5000 and prints a message specifying the listening port as soon as it starts. This message is crucial for clarifying failing tests, which previously only displayed "**timeout: Did you write an infinite loop?**"

MSilva95 (approved):
LGTM!

PR #2430, opened 2024-02-01
Same Ques was repeated twice in README.md

sagarishere:
As the same questionw as repeated twice, I just made it shorter.

DEV-#### | SUP-#### | CON-####

### Why?

### Solution Overview

### Implementation Details

### Build Images

>- `💻 Sh` 
>- `🚀 JS`
>- `🧩 DOM`


nprimo (approved):
@sagarishere, thank you for your contribution!

PR #2452, opened 2024-02-20
fix music-on subject issue

zanninso:
The subject contains technical terminology that has not been adequately defined, leading to potential confusion among students regarding the expected outputs of their work.


MSilva95 (approved):
LGTM!

PR #1501, opened 2022-10-11
Update README.md

sagarishere:
escaping !

PR #2477, opened 2024-03-01
CON-2477 update-example-command-in 1ascii-art-output`

nprimo:
- substitute " with ' to avoid special symbol interpretation

Addresses the issue raised [here](https://github.com/01-edu/public/pull/1501)


MSilva95 (approved):
LGTM!

PR #2483, opened 2024-03-05
CON-2399-keep-naming-convention

zamazzal:
### Why?
Keep naming convention

### Solution Overview
change name of cloud press project


Oumaimafisaoui (approved):
Changed the case type from Camel Case to Kebab Case.

PR #2492, opened 2024-03-11
CON-2496 fix broken link in `play-with-containers` audit

nprimo:
- Fix small format issue.
- Restructure audit question to have a question right after instruction.


MSilva95 (approved):
LGTM!

PR #2510, opened 2024-03-26
CON-2526 add new GA to check for broken links

nprimo:
Before merging:
- be sure that the new label to skip check is added 
- consider moving the GA to 01-edu


Frenchris (approved):
LGTM. the "uses: nprimo/check-links@v0.1" will need to be forked or changed to our organization after tests of the script are satisfactory.

Frenchris (changes requested):
It just needs the change to point to 01-edu/check-links

Frenchris (approved):
LGTM! :)

PR #2517, opened 2024-04-08
CON-2619 piscine-ux audit for analytics

MSilva95:
CON-2619

### Why?

Wrong audit, it was the same as the previous exercise

### Solution Overview

Create new audit.


nprimo (approved):
LGTM :rocket:

PR #2529, opened 2024-04-09
CON-2565-Markdown-Palindrom-exercise

zanninso:


subjects/java/checkpoints/palindrom/README.md
```diff
@@ -0,0 +1,38 @@
+## Palindrome
+
+### Instructions
+
+In a file named `Palindrome.java` write a function `isPalindrome` that returns true if the String in a parameter is a palindrome, i.e. Have the same spelling backward and forwards (e.g. 'kayak').
```
nprimo, line 5: `Have` should be all lower case

nprimo (changes requested):
This exercise is already present (`palindrome`) in the piscine - I wouldn't use the same exercises for exam and piscine. If this is the case, there is no need to create a new subject

subjects/java/checkpoints/palindrom/README.md
zanninso: The comparison will be case sensitive

zanninso:
> This exercise is already present (`palindrome`) in the piscine - I wouldn't use the same exercises for exam and piscine. If this is the case, there is no need to create a new subject

yes but there is a small difference between the two in the exam they shouldn't provide the solution inside a folder and i see is better to separate between the piscine exercises and the exam exercises

subjects/java/checkpoints/almost-palindrom/README.md
```diff
@@ -0,0 +1,39 @@
+## Palindrome
+
+### Instructions
+
+In a file named `AlmostPalindrome.java` write a function `isAlmostPalindrome` that returns true if the String in parameter is almost palindrome.
+A word is considered "almost a palindrome" if removing one letter from the word makes it a palindrome.
```
nprimo, line 6: What should happen for edge cases? Is an empty string "almost palindrome"?

nprimo, line 6: Is the check case insensitive or cases counts? I would provide some example with capital letters or add some info in the subject to avoid confusion

nprimo (changes requested):

subjects/java/checkpoints/almost-palindrom/README.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+### Usage
+
+Here is a possible `ExerciseRunner.java` to test your function :
+
+```java
+public class ExerciseRunner {
+    public static void main(String[] args) {
+        System.out.println(AlmostPalindrome.isAlmostPalindrome("reviver"));
```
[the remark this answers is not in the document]
zanninso: yes you are right it was a mistake, thanks for notice

subjects/java/checkpoints/almost-palindrom/README.md
zanninso: i changed the test to clarify that

zanninso: They will get just string formed with at least 3 letters, i will add that.

subjects/java/checkpoints/almost-palindrom/README.md
```diff
@@ -0,0 +1,41 @@
+## AlmostPalindrome
+
+### Instructions
+
+In a file named `AlmostPalindrome.java` write a function `isAlmostPalindrome` that returns true if the String in parameter is almost palindrome.
+A word is considered "almost a palindrome" if removing one letter from the word makes it a palindrome.
+
+> 💡 The String in parametre will contains at least 3 letters.
```
nprimo, line 8: There are 2 small grammar mistakes:
- `will contains` -> should be `will contain`
- `parametre` -> should be `parameter`. An alternative could also be `The input String will contain...`

nprimo (changes requested):
Just 2 small grammar fixes. Additionally, I would add maybe one sentence as well to specify that the palindrome check is case insensitive. With the new test at least it's possible to understand that so up to you :)

subjects/java/checkpoints/almost-palindrom/README.md
zanninso: done

nprimo (approved):
LGTM :rocket:

PR #2530, opened 2024-04-09
CON-2569-Markdown-Factorial-exercise

zanninso:


subjects/java/checkpoints/factorial/README.md
```diff
@@ -0,0 +1,38 @@
+## Factorial
+
+### Instructions
+
+In a file named `Factorial.java` write a function `factorial` that calculates the factorial of a non-negative integer given as a parameter.
+
+> The factorial of a non-negative integer n (denoted as n!) is the product of all positive integers less than or equal to n. For example, 5! = 5 × 4 × 3 × 2 × 1 = 120.
+
```
nprimo, line 8: What should happen if the function receives a negative number?
What is the expected result for `0!`?

nprimo (changes requested):

subjects/java/checkpoints/factorial/README.md
zanninso: They will not receive a negative number because i asked them to calculate the factorial of non negative number,
The factorial of 0 is 1 i guess this is a knowing information. and doesn't need further explanation

nprimo: I would add the information about the 0! since that's something you either know or don't (it's not about logic) 

> The value of 0! is 1, according to the convention for an [empty product](https://en.wikipedia.org/wiki/Empty_product).[[1]](https://en.wikipedia.org/wiki/Factorial#cite_note-gkp-1)

nprimo (changes requested):

subjects/java/checkpoints/factorial/README.md
zanninso: okay i will add it then

nprimo (approved):
LGTM

PR #2548, opened 2024-04-22
CON-2626 new exercise findspairs

MSilva95:
CON-2626

### DIFF
--- a/subjects/findpairs/README.md
+++ b/subjects/findpairs/README.md
@@ -1,4 +1,4 @@
-## Find Pairs with Sum
+## Find Pairs
 
 ### Instructions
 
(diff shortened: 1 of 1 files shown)


subjects/findpairs/README.md
```diff
@@ -0,0 +1,32 @@
...
+Write a program that finds all pairs of elements in an integer array that sum up to a given target value. The program should output a list of pairs, each representing the indices of the elements that form the pair.
+
+In this exercise you must take in consideration the following:
+
+- Ensure its possible to have positive or negative integers in the array
+- Ensure each element is used only once in a pair
+- Allow for multiple pairs to sum up to the target value
+- Return the message "No pairs found." when no pair is present
+- Return the message "Invalid target sum." if the target is invalid
+- Return the message "Invalid number." if the number in the array is invalid
```
nprimo, line 14: What about a wrong format that is not included in the above? For example:
```console
go run . "1, 2, 3, 4" "2"
```
or 
```console
go run . "[1, 2, 3, 4" "5"
```
Is this a valid input or not?

nprimo (changes requested):
I would add some extra information about "other invalid input" cases not specified in the options

subjects/findpairs/README.md
MSilva95: added some new info in the subject and test for this cases

nprimo (approved):
LGTM :rocket:

PR #2563, opened 2024-05-15
CON-2695 update game dev projects' subjects and audits

nprimo:
CON-2695

### Why?

Game Dev assets might be larger than the file maximum size allowed for our schools' Gitea 

### Solution Overview

Update subjects and audit to allow learners to use GitHub for large files


MSilva95 (approved):
LGTM!

PR #2564, opened 2024-05-15
CON-2696 Update `quadchecker` subject

nprimo:
CON-2696

### Why?

Update `quadchecker` subject according to https://github.com/01-edu/go-tests/pull/185


MSilva95 (approved):
LGTM!

PR #2569, opened 2024-05-20
 CON-2699  Change the cmd 'tar -czf home.tar.gz .' to 'ls -lRr / 2>1 >/dev/null'

zanninso:
change the cmd 'tar -czf home.tar.gz .' to 'ls -lRr / 2>1 >/dev/null' to avoid failing of the first cmd because of the versions differences


MSilva95 (approved):
LGTM!

PR #2578, opened 2024-05-21
CON-2664 [CHECKPOINT-RESTRUCTURE] Update `lastword` to be a function instead of a program

nprimo:
[CON-2663](https://01talent.atlassian.net/browse/CON-2663)

[CON-2663]: https://01talent.atlassian.net/browse/CON-2663?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/subjects/lastword/README.md
+++ b/subjects/lastword/README.md
@@ -2,24 +2,48 @@
 
 ### Instructions
 
-Write a program that takes a `string` and displays its last word, followed by a newline (`'\n'`).
+Write a function `LastWord that takes a `string`and returns its last word followed by a`\n`.
 
 - A word is a section of `string` delimited by spaces or by the start/end of the `string`.
 
-- The output will be followed by a newline (`'\n'`).
+### Expected function
 
-- If the number of arguments is different from 1, or if there are no word, the program displays nothing.
+```go
+func LastWord(s string) string{
+
+}
+```
 
 ### Usage
 
+Here is a possible program to test your function :
+
+```go
+package main
+
+import (
+	"piscine"
+
+	"github.com/01-edu/z01"
+)
+
+func main() {
+	z01.PrintRune(piscine.LastWord("this        ...       is sparta, then again, maybe    not"))
+	z01.PrintRune(piscine.LastWord(" "))
+	z01.PrintRune(piscine.LastWord(" lorem,ipsum "))
+}
+```
+
+And its output :
+
 ```console
-$ go run . "FOR PONY" | cat -e
-PONY$
-$ go run . "this        ...       is sparta, then again, maybe    not" | cat -e
+$ go run . | cat -e
 not$
-$ go run . "  "
-$ go run . "a" "b"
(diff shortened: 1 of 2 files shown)


subjects/lastword/README.md
```diff
@@ -2,24 +2,48 @@
...
 ```console
-$ go run . "FOR PONY" | cat -e
-PONY$
-$ go run . "this        ...       is sparta, then again, maybe    not" | cat -e
+$ go run . | cat -e
 not$
-$ go run . "  "
-$ go run . "a" "b"
-$ go run . "  lorem,ipsum  " | cat -e
+$
```
MSilva95, line 42: I think the students may think that this is the same as the terminal prompt.
This happened in the past, maybe we can remove the cat -e flag or move the empty example to the last position so that its a bit more clear.

subjects/lastword/README.md
```diff
@@ -2,24 +2,48 @@
...
+
+import (
+	"piscine"
+
+	"github.com/01-edu/z01"
+)
+
+func main() {
+	z01.PrintRune(piscine.LastWord("this        ...       is sparta, then again, maybe    not"))
+	z01.PrintRune(piscine.LastWord(" "))
```
MSilva95, line 32: bellow i'm talking about this, to move this one to the bottom

subjects/lastword/README.md
```diff
@@ -2,24 +2,48 @@
...
+package main
+
+import (
+	"piscine"
+
+	"github.com/01-edu/z01"
+)
+
+func main() {
+	z01.PrintRune(piscine.LastWord("this        ...       is sparta, then again, maybe    not"))
```
MSilva95, line 31: here you can use "fmt" if you want it to be more clean, otherwise just close this conv its fine like this as well

MSilva95 (changes requested):
Requested a small change to the output to avoid confusion

subjects/lastword/README.md
nprimo: I'll change it with `fmt.Print`

subjects/lastword/README.md
nprimo: I guess they should be able to count how many new lines are there, but I'll change it to reduce future complaints :sweat_smile:

nprimo: I guess they should be able to count how many new lines should be there, but I'll change it to reduce future complaints :sweat_smile:

MSilva95: yeah, but it happened a couple of times already, an i bet it will happen some more 😅

MSilva95 (approved):
MERGE ONLY AFTER THE MIGRATION!!!
LGTM!

HarryVasanth (approved):
> [!NOTE]
>  ✅ Production and pre-production prerequisites have been deployed.
> #### Clear to merge!

PR #2615, opened 2024-06-27
CON-2833-Markdown-Distinct-Substring-Lenght-Exercise

zanninso:
[CON-2833](https://01talent.atlassian.net/browse/CON-2833)

[CON-2833]: https://01talent.atlassian.net/browse/CON-2833?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


subjects/java/checkpoints/distinct-substring-lenght/README.md
```diff
@@ -0,0 +1,45 @@
+## Distinct Substring length
+
+### Instructions
+
+Create a class `DistinctSubstringLength` that provides a method to find the length of the longest substring without repeating characters in a given string.
```
nprimo, line 5: What if a string is made of all unique char? Is the length of the "longest substring without repeating characters" supposed to be the length of the string?

nprimo, line 5: I would specify if this is case-sensitive or not. In other exercises, the checks where case insensitive while here it seems to be case sensitive (from the tests)

nprimo, line 5: I would add a definition of "substring": I only understood this from one of the tests.

For example:
>... a substring is a contiguous sequence of [characters](https://en.wikipedia.org/wiki/Character_(computing)) within a [string](https://en.wikipedia.org/wiki/String_(computer_science)).

From [Wikipedia](https://en.wikipedia.org/wiki/Substring)

nprimo (changes requested):

subjects/java/checkpoints/distinct-substring-lenght/README.md
zanninso: yes~

zanninso: i do it in purpose, not all the informations should be clearly defined in the subject the have to get the informations from the example too.

zanninso: the default is to be case-sensitive, but there is no harm to specify that :)

subjects/java/checkpoints/distinct-substring-lenght/README.md
```diff
@@ -2,7 +2,7 @@
 
 ### Instructions
 
-Create a class `DistinctSubstringLength` that provides a method to find the length of the longest substring without repeating characters in a given string.
+Create a class `DistinctSubstringLength` that provides a method to find the length of the longest substring without repeating characters in case sensitive withen the given string.
```
nprimo, line 5: I would keep the first version of the sentence and add a short additional sentence (or an example) to show that the distinctions between characters should be case-sensitive.

An example of a sentence to be added:
```
The distinction between characters should be case-sensitive.
```

nprimo (changes requested):

subjects/java/checkpoints/distinct-substring-lenght/README.md
zanninso: done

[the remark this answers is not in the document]
zanninso: done

PR #2617, opened 2024-06-27
CON-2827 Markdown Is Anagram Exercise

zanninso:
[CON-2827](https://01talent.atlassian.net/browse/CON-2827)

[CON-2827]: https://01talent.atlassian.net/browse/CON-2827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/subjects/java/checkpoints/is-anagram/README.md
+++ b/subjects/java/checkpoints/is-anagram/README.md
@@ -24,10 +24,10 @@ public class ExerciseRunner {
         AnagramChecker checker = new AnagramChecker();
 
         // Test cases
-        System.out.println(checker.isAnagram("listen", "silent")); // Expected output: true
-        System.out.println(checker.isAnagram("triangle", "integral")); // Expected output: true
-        System.out.println(checker.isAnagram("apple", "pale")); // Expected output: false
-        System.out.println(checker.isAnagram("Astronomer", "Moon starer")); // Expected output: true
+        System.out.println(checker.isAnagram("listen", "silent"));
+        System.out.println(checker.isAnagram("triangle", "integral"));
+        System.out.println(checker.isAnagram("apple", "pale"));
+        System.out.println(checker.isAnagram("Astronomer", "Moon starer"));
     }
 }
 ```
@@ -40,6 +40,6 @@ $ java -cp build ExerciseRunner
 true
 true
 false
-true
+false
 $
 ```
--- a/subjects/java/checkpoints/is-anagram/README.md
+++ b/subjects/java/checkpoints/is-anagram/README.md
@@ -2,7 +2,7 @@
 
 ### Instructions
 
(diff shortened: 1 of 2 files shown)


subjects/java/checkpoints/is-anagram/README.md
```diff
@@ -0,0 +1,45 @@
+## Is Anagram
+
+### Instructions
+
+Create a class `AnagramChecker` that provides a method to check if two strings are anagrams. An anagram is a word or phrase formed by rearranging the letters of a different word or phrase, typically using all the original letters exactly once.
```
nprimo, line 5: Is the check going to be case-sensitive?
It might be worth saying that two empty strings should be considered anagrams of each other

subjects/java/checkpoints/is-anagram/README.md
```diff
@@ -0,0 +1,45 @@
...
+public class AnagramChecker {
+    public boolean isAnagram(String str1, String str2) {
+        // Implementation to check if str1 and str2 are anagrams
+    }
+}
+```
+
+### Usage
+
+Here is a possible ExerciseRunner.java to test your class:
```
nprimo, line 19: A minor comment: it would be nice to have `ExerciseRunner.java` formatted as the other subjects

nprimo (changes requested):

subjects/java/checkpoints/is-anagram/README.md
zanninso: no need to specify the case of empty string because it's logically correct if think of it, then they need to know it by them selves.

subjects/java/checkpoints/is-anagram/README.md
zanninso: done

subjects/java/checkpoints/is-anagram/README.md
```diff
@@ -0,0 +1,45 @@
...
+```java
+public class ExerciseRunner {
+    public static void main(String[] args) {
+        AnagramChecker checker = new AnagramChecker();
+
+        // Test cases
+        System.out.println(checker.isAnagram("listen", "silent")); // Expected output: true
+        System.out.println(checker.isAnagram("triangle", "integral")); // Expected output: true
+        System.out.println(checker.isAnagram("apple", "pale")); // Expected output: false
+        System.out.println(checker.isAnagram("Astronomer", "Moon starer")); // Expected output: true
```
nprimo, line 30: The result in the test for this case is `false` - from the definition (considering also space as chars) it should be false, correct?

nprimo (changes requested):

subjects/java/checkpoints/is-anagram/README.md
zanninso: yes, its fixed

nprimo:
@zanninso please address the open comment [here](https://github.com/01-edu/java-tests/pull/50)

nprimo (approved):
LGTM

PR #2620, opened 2024-07-01
CON-2869 Markdown Age Finder

zanninso:
[CON-2869](https://01talent.atlassian.net/browse/CON-2869)

[CON-2869]: https://01talent.atlassian.net/browse/CON-2869?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/subjects/java/checkpoints/age-finder/README.md
+++ b/subjects/java/checkpoints/age-finder/README.md
@@ -1,0 +1,56 @@
+## Age Finder
+
+### Instructions
+
+Create a class `AgeFinder` that provides a method to calculate the age from a given date. The date will be provided in the format `yyyy-MM-dd`.
+
+In case of any error the method `calculateAge` should return `-1`
+
+### Expected Class
+
+```java
+import java.time.LocalDate;
+import java.time.Period;
+import java.time.format.DateTimeFormatter;
+
+public class AgeFinder {
+    public int calculateAge(String date) {
+        // Implementation to calculate the age from the given date
+    }
+}
+```
+
+### Usage
+
+Here is a possible `ExerciseRunner.java` to test your class:
+
+```java
+public class ExerciseRunner {
+    public static void main(String[] args) {
+        AgeFinder AgeFinder = new AgeFinder();
+
+        // Test case 1
+        String date1 = "2000-01-01";
+        System.out.println("Age: " + AgeFinder.calculateAge(date1));
+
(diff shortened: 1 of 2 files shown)


subjects/java/checkpoints/age-finder/README.md
```diff
@@ -0,0 +1,56 @@
...
+        // Test case 3
+        String date3 = "2010-12-25";
+        System.out.println("Age: " + AgeFinder.calculateAge(date3));
+    }
+}
+```
+
+### Expected Output
+
+```shell
```
nprimo, line 49: To avoid having to update this command on a yearly basis I will suggest to add the command `date` with its output at the beginning :)

subjects/java/checkpoints/age-finder/README.md
```diff
@@ -0,0 +1,56 @@
+## Age Finder
+
+### Instructions
+
+Create a class `AgeFinder` that provides a method to calculate the age from a given date. The date will be provided in the format `yyyy-MM-dd`.
+
+In case of any error the method `calculateAge` should return `-1`
```
nprimo, line 7: I would either specify or provide an example to show that dates in the future have to be considered errors

nprimo (changes requested):

subjects/java/checkpoints/age-finder/README.md
zanninso: yeah, i forget about that, thanks

subjects/java/checkpoints/age-finder/README.md
zanninso: I added a hint to that.

subjects/java/checkpoints/age-finder/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+> 💡 Be careful with the future.
```
nprimo, line 9: This doesn't say a lot :sweat_smile: It's not clear what should happen since the `int` return from the function can be negative (even if probably it won't make sense). I would add, more than "Be careful with the future", something like "Age can only be a positive number or zero" (I don't know about 0)

nprimo (changes requested):

PR #2621, opened 2024-07-01
CON-2875 Markdown Monthly Period

zanninso:
[CON-2875](https://01talent.atlassian.net/browse/CON-2875)

[CON-2875]: https://01talent.atlassian.net/browse/CON-2875?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/subjects/java/checkpoints/monthly-period/README.md
+++ b/subjects/java/checkpoints/monthly-period/README.md
@@ -1,0 +1,60 @@
+## Monthly Period
+
+### Instructions
+
+Create a class `MonthlyPeriod` that provides a method to calculate the period between two given dates in terms of months and years. The dates will be provided in the format `yyyy-MM-dd`.
+
+In case of any error the method `calculatePeriod` should return `Error`
+
+> 💡 Going to the past is possible!
+
+### Expected Class
+
+```java
+import java.time.LocalDate;
+import java.time.Period;
+import java.time.format.DateTimeFormatter;
+
+public class MonthlyPeriod {
+    public String calculatePeriod(String startDate, String endDate) {
+        // Implementation to calculate the period between two dates in months and years
+    }
+}
+```
+### Usage
+
+Here is a possible `ExerciseRunner.java` to test your class:
+
+```java
+public class ExerciseRunner {
+    public static void main(String[] args) {
+        MonthlyPeriod MonthlyPeriod = new MonthlyPeriod();
+
+        // Test case 1
+        String startDate1 = "2020-01-01";
+        String endDate1 = "2023-06-15";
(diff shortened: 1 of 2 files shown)


subjects/java/checkpoints/day-of-week/ExerciseRunner.java
nprimo, line 1: This seems to be an extra file from another exercise / different ticket

subjects/java/checkpoints/day-of-week/README.md
nprimo, line 1: This seems to be an extra file form another exercise / different ticket

subjects/java/checkpoints/monthly-period/README.md
```diff
@@ -0,0 +1,60 @@
+## Monthly Period
+
+### Instructions
+
+Create a class `MonthlyPeriod` that provides a method to calculate the period between two given dates in terms of months and years. The dates will be provided in the format `yyyy-MM-dd`.
```
nprimo, line 5: What is the expected format of the period?
I deduce from the example that it should be `x years and y months` so can the numbers of days be omitted?
Should the `years` be `year` (same for `months`) if it's only 1?

nprimo (changes requested):

subjects/java/checkpoints/monthly-period/README.md
nprimo, line 1: This file needs to be formatted

subjects/java/checkpoints/monthly-period/README.md
zanninso: The expected format is clearly showed in the examples. I'm agree with you if there is only 1 it should be `month`/`year`.

nprimo: What about the days? I see you're not supposed to return the days, but what if the difference is 1 year, 2 months and 30 days: is it supposed to be 1 year and 2 months or 1 year and 3 months (since it's closer to being 3 months than 2 months)? 
I feel adding an example or changing the first one by pushing the `endDate` of the first one to `2023-06-30` should already give enough information on how these cases should be handled

nprimo (changes requested):

subjects/java/checkpoints/monthly-period/README.md
zanninso: the diff should be done using the standard library, so they will always get the right period.
here i asked just for months and years while the month is not completed they can't count it as a fulll month

nprimo (approved):

PR #2632, opened 2024-07-03
CON-2902-Markdown-Top-Frequents-Exercise

zanninso:
[CON-2902](https://01talent.atlassian.net/browse/CON-2902)

[CON-2902]: https://01talent.atlassian.net/browse/CON-2902?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/subjects/java/checkpoints/top-frequents/README.md
+++ b/subjects/java/checkpoints/top-frequents/README.md
@@ -1,0 +1,55 @@
+## Top Frequents
+
+### Instructions
+
+Given a non-empty array of integers, return the k most frequent elements. Use a hash map and a heap for efficient lookup and retrieval.
+
+### Expected Class
+
+```java
+import java.util.*;
+
+public class TopFrequents {
+    public List<Integer> findTopKFrequent(int[] nums, int k) {
+        // Implementation to find the k most frequent elements
+    }
+}
+```
+
+### Usage
+
+Here is a possible `ExerciseRunner.java` to test your class:
+
+```java
+public class ExerciseRunner {
+    public static void main(String[] args) {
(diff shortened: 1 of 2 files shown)


subjects/java/checkpoints/top-frequents/README.md
```diff
@@ -0,0 +1,55 @@
+## Top Frequents
+
+### Instructions
+
+Given a non-empty array of integers, return the k most frequent elements. Use a hash map and a heap for efficient lookup and retrieval.
```
nprimo, line 5: I would add what is the sorting logic for the returned array: is it supposed to be ordered by "appearance" (smallest index in "original" array = smallest index in the returned array) or should it be based on "standard" ascending sorting of integer?

From the examples, both logics could be valid.

nprimo, line 5: What if k is greater than the number of unique elements in the array? Maybe adding an example to show what's expected could be enough

nprimo (changes requested):

subjects/java/checkpoints/top-frequents/README.md
zanninso: it should be ordred by appearance first then if there is there is frequent equality the one that have the smallest index in the original array will be first.

zanninso: okay i will add a test for it

nprimo (approved):
LGTM :rocket:

PR #2634, opened 2024-07-03
CON-2896-Markdown-Longest-Common-Prefix-Exercise

zanninso:
[CON-2896](https://01talent.atlassian.net/browse/CON-2896)

### DIFF
--- a/subjects/java/checkpoints/longest-common-prefix/README.md
+++ b/subjects/java/checkpoints/longest-common-prefix/README.md
@@ -1,0 +1,50 @@
+## Longest Common Prefix
+
+### Instructions
+
+Write a function to find the longest common prefix string amongst an array of strings. If there is no common prefix, return an empty string `""`.
+
+### Expected Class
+
+```java
+public class LongestCommonPrefix {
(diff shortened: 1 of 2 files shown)


subjects/java/checkpoints/longest-common-prefix/README.md
```diff
@@ -0,0 +1,50 @@
+## Longest Common Prefix
+
+### Instructions
+
+Write a function to find the longest common prefix string amongst an array of strings. If there is no common prefix, return an empty string `""`.
```
nprimo, line 5: Nothing major, but I would specify if the prefix is supposed to be case-sensitive or not. In other exercises the comparison was not case-insensitive while here it is case-sensitive (I only found out looking at the tests)

nprimo (changes requested):

subjects/java/checkpoints/longest-common-prefix/README.md
zanninso: yes, you are right i forget to add that.
its fixed now thanks

nprimo (approved):
LGTM

PR #2641, opened 2024-07-08
CON-2774 Update DOM Dockerfile

lotoussa:
[CON-2774](https://01talent.atlassian.net/browse/CON-2774)

### Why?
Update the DOM Dockerfile. This version was used in previous Discovery piscine in order to properly work, and is now push to the public official repository as it will be used for the IA Discovery Piscine.

### Solution Overview
The Dockerfile content come from a copy paste of another docker file of the public repository, it has been tested already on previous Discovery Piscine and seem to work but still need a double check.

[CON-2774]: https://01talent.atlassian.net/browse/CON-2774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dom/Dockerfile
+++ b/dom/Dockerfile
@@ -1,6 +1,24 @@
-FROM buildkite/puppeteer:7.1.0
+FROM alpine:latest
 
 WORKDIR /app
-COPY dom .
+COPY . .
 COPY subjects ./subjects
+#RUN npm i
+RUN apk add --no-cache \
+    chromium \
+    nss \
+    freetype \
+    harfbuzz \
+    ca-certificates \
+    ttf-freefont \
+    nodejs \
+    yarn
+
+# Tell Puppeteer to skip installing Chrome. We'll be using the installed package.
+ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true \
+    PUPPETEER_EXECUTABLE_PATH=/usr/bin/chromium-browser
+RUN apk add --update nodejs npm
+RUN npm i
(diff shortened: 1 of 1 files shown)


nprimo (commented):
@lotoussa  I left just some general comments. Could you please elaborate a bit on why you want to make these changes?

Besides that, this image is also used for part of our content for the JS part of the curriculum besides the Discovery Piscine (just for your information) :)

dom/Dockerfile
```diff
@@ -1,6 +1,24 @@
-FROM buildkite/puppeteer:7.1.0
+FROM alpine:latest
 
 WORKDIR /app
-COPY dom .
```
nprimo, line 4: Keep in mind that his `Dockerfile` needs to work for our GitHub Actions :) 
The command used to build the image is [here](https://github.com/01-edu/public/blob/b863e07340cb2af393bd9036a1a338b147b6f56f/.github/workflows/ga-image-build-branch.yml#L45)

lfoussat:
> @lfoussat I left just some general comments. Could you please elaborate a bit on why you want to make these changes?
> 
> Besides that, this image is also used for part of our content for the JS part of the curriculum besides the Discovery Piscine (just for your information) :)

I guess you meant @lotoussa and not @lfoussat ?

lotoussa:
@nprimo hello yes i believe it's not relevant

PR #2647, opened 2024-07-17
CON-2951-Tests-DPxAI-Quest00-Ex02

Oumaimafisaoui:
https://01talent.atlassian.net/browse/CON-2951

### DIFF
--- a/dom/select-then-style_test.js
+++ b/dom/select-then-style_test.js
@@ -1,0 +1,49 @@
+
+export const tests = []
+
+tests.push(async ({ eq }) => {
+  // check the CSS stylesheet is linked in the head tag
+
+  await eq.$('head link', {
+    rel: 'stylesheet',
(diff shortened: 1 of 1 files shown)


dom/select-then-style_test.js
```diff
@@ -0,0 +1,49 @@
+
+export const tests = []
+
+tests.push(async ({ eq }) => {
+  // check the CSS stylesheet is linked in the head tag
+
+  await eq.$('head link', {
+    rel: 'stylesheet',
+    href: 'http://localhost:9898/select-and-style/select-and-style.css',
```
nprimo, line 9: should this be `select-then-style` instead of `select-and-style`?

nprimo (changes requested):

dom/select-then-style_test.js
Oumaimafisaoui: Adressed.

nprimo (approved):
LGTM

PR #2659, opened 2024-07-24
CON-3015 Update `printif` and `printifnot` examples to be coherent with subject

nprimo:
CON-3015
SUP-1192 - issues 2 and 3


MSilva95 (approved):
LGTM!

PR #2667, opened 2024-07-28
CON-3006 Subject for ``Quest01/play-with-variables``

Oumaimafisaoui:
Resolving: [CON-3006]('https://01talent.atlassian.net/browse/CON-3006')

[CON-3006]: https://01talent.atlassian.net/browse/CON-3006?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2672, opened 2024-07-31
Tests(DPxAI): Add test for Quest01 play-with-variables

lotoussa:
[CON-3007](https://01talent.atlassian.net/browse/CON-3007)

### DIFF
--- a/js/tests/play-with-variables_test.json
+++ b/js/tests/play-with-variables_test.json
@@ -1,0 +1,22 @@
+[
+  {
+    "description": "escapeFromDelimiters is declared and includes a double-quote",
+    "code": "if (typeof escapeFromDelimiters === 'undefined') {\n  throw Error(\n    `You didn't even define the variable... we've been through this already !`,\n  )\n}\n\nif (!escapeFromDelimiters.includes('\"')) {\n  throw Error('escapeFromDelimiters must include a double-quote\"')\n}"
+  },
+  {
+    "description": "escapeFromDelimiters includes a single-quote",
+    "code": "if (!escapeFromDelimiters.includes(\"'\")) {\n  throw Error(\"escapeFromDelimiters must include a single-quote'\")\n}"
+  },
+  {
+    "description": "escapeFromDelimiters includes a backtick",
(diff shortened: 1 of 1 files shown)


js/tests/play-with-variables_test.json
nprimo, line 1: The file name needs to be updated to `play-with-variables.json` to work as expected.

Is there any reason to have the test as `json` file instead of `js`?

nprimo (changes requested):
Test fails with the solution in 01-edu/solutions#46 wit the following output
```
escapeFromDelimiters is declared and includes a double-quote: FAIL

======= Error ======
 -> power is not defined


======= Code =======
// Your code
let escapeFromDelimiters = "This is an example of all three quotes: \`, \", and \'"
let escapeTheEscape = "This is an example of a backslash: \\"
power = "levelMax";

// The tests
if (typeof escapeFromDelimiters === 'undefined') {
  throw Error(
    `You didn't even define the variable... we've been through this already !`,
  )
}

if (!escapeFromDelimiters.includes('"')) {
  throw Error('escapeFromDelimiters must include a double-quote"')
};
```

js/tests/play-with-variables_test.json
```diff
@@ -0,0 +1,22 @@
...
+    "description": "escapeFromDelimiters includes a backtick",
+    "code": "if (!escapeFromDelimiters.includes('`')) {\n  throw Error('escapeFromDelimiters must include a backtick `')\n}"
+  },
+  {
+    "description": "escapeTheEscape includes a backslash",
+    "code": "if (!new TextEncoder().encode(escapeTheEscape).includes(92)) {\n  throw Error('escapeTheEscape must includes a backslash')\n}"
+  },
+  {
+    "description": "The value of power must have changed",
+    "code": "let power = 'under 9000'\n\n// Your code\n\nequal(love, 'levelMax')"
```
lotoussa, line 20: https://github.com/01-edu/public/pull/2667#discussion_r1698546657

also about this test i have a question here please @nprimo

nprimo:
@lotoussa the current tests are failing the current solution here 01-edu/solutions#48

nprimo (approved):

Oumaimafisaoui (approved):
LGTM

PR #2679, opened 2024-08-08
CON-3045 Tests(DPxAI): Upload test for quest01 objects-around

lotoussa:
`https://01talent.atlassian.net/browse/CON-3045`

>  -> the two last tests cannot work because the test declare the object called robot and the solution declare it also. So the subject should probably ask to search for name age and hasEnergy in another variable name than robot to avoid conflict.
Fixed

### DIFF
--- a/js/tests/objects-around.json
+++ b/js/tests/objects-around.json
@@ -1,19 +1,19 @@
 [
   {
-    "description": "variable robot is declared and of type object",
-    "code": "equal(typeof robot, 'object')"
+    "description": "variable myRobot is declared and of type object",
+    "code": "let robot = {\n  name: 'Freddy',\n  age: 27,\n  hasEnergy: false,\n}\n\n// Your code\n\nequal(typeof myRobot, 'object')"
   },
   {
-    "description": "property name from robot is of type string",
-    "code": "equal(typeof robot.name, 'string')"
+    "description": "property name from myRobot is of type string",
+    "code": "let robot = {\n  name: 'Freddy',\n  age: 27,\n  hasEnergy: false,\n}\n\n// Your code\n\nequal(typeof myRobot.name, 'string')"
   },
   {
-    "description": "property age from robot is of type number",
-    "code": "equal(typeof robot.age, 'number')"
+    "description": "property age from myRobot is of type number",
+    "code": "let robot = {\n  name: 'Freddy',\n  age: 27,\n  hasEnergy: false,\n}\n\n// Your code\n\nequal(typeof myRobot.age, 'number')"
   },
   {
-    "description": "property hasEnergy from robot is of type boolean",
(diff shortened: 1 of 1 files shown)


js/tests/objects-around.json
```diff
@@ -0,0 +1,26 @@
...
+    "code": "equal(typeof robot.name, 'string')"
+  },
+  {
+    "description": "property age from robot is of type number",
+    "code": "equal(typeof robot.age, 'number')"
+  },
+  {
+    "description": "property hasEnergy from robot is of type boolean",
+    "code": "equal(typeof robot.hasEnergy, 'boolean')"
+  },
```
Oumaimafisaoui, line 17: I changed the name of the object from robot to MyRobot

Oumaimafisaoui (changes requested):
Please follow the new naming.

nprimo (changes requested):
The tests fail the provided solution here 01-edu/solutions#50 (even after the suggested change)
```
variable myRobot is declared and of type object: PASS
property name from myRobot is of type string: PASS
property age from myRobot is of type number: PASS
property hasEnergy from myRobot is of type boolean: PASS
all 3 variable should be defined and have the right values: FAIL

======= Error ======
 -> Expected values to be strictly deep-equal:
+ actual - expected

  {
+   age: 5,
+   hasEnergy: true,
+   name: 'RoboMax'
-   age: 27,
-   hasEnergy: false,
-   name: 'Freddy'
  }


======= Code =======
// Provided setup
let robot = {
  name: 'Freddy',
  age: 27,
  hasEnergy: false,
}

// Your code
let myRobot = {
  name: 'RoboMax',
  age: 5,
  hasEnergy: true,
}

let name = myRobot.name
let age = myRobot.age
let hasEnergy = myRobot.hasEnergy;

// The tests
equal({ name, age, hasEnergy }, robot);

```

lotoussa:
@nprimo the solution PR is not updated yet, i will push a commit with the correct version
the failed solution you showed us doesn't fit the subject i think. You declare myRobot, and then you select value from myRobot but we should select value from robot.
it's 2 independant tasks.
if you think it's confusing we can think of other variable name.

exemple of the code i'll push to the solution file :
```
let myRobot = {
    name: "RoboMax",
    age: 5,
    hasEnergy: true,
};

let name = robot.name;
let age = robot.age;
let hasEnergy = robot.hasEnergy;

variable myRobot is declared and of type object: PASS
property name from myRobot is of type string: PASS
property age from myRobot is of type number: PASS
property hasEnergy from myRobot is of type boolean: PASS
all 3 variable should be defined and have the right values: PASS
value should also work for Jean-Pierre: PASS
```

nprimo (approved):

PR #2725, opened 2024-09-19
CON-3139 Game-dev introduction project

EldritchGriffin:
CON-3139

This project involves creating a simple 2.5D side-scroller game designed to introduce the student to the fundamentals of Unreal Engine 5 and game development, without the complexity of advanced mechanics. It focuses on the essential skills and concepts they need to kickstart their gamedev journey.


subjects/gaming/2.5d-adventure/audit/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+###### Does the collectible play a sound when collected?
+
+#### HUD
+
+###### Does the HUD widget display a value related to the `collectible`?
+
+###### Is the HUD updated each time a `collectible` is picked up?
+
+###### Is the HUD also updated when picking a `collectible` that spawned from killing an enemy?
```
MSilva95, line 45: This part is not mentioned in the subject, we should add it otherwise the student dont know about it

subjects/gaming/2.5d-adventure/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+The game loop consists of a spawn point where the player `starts` and `respawns` when dead.
+And an end point that defines the player goal to finish the level.
+When the player reaches the end point:
+
+- a menu with options to `restart` or `quit` the game should be displayed.
+
+#### Level design
+
+Your level design shouldn't consist of just a long, empty run to the finish point. However, since creating a level is a creative process, you have the freedom to design it as you see fit. For example, you can include moving platforms or skill-based platforming elements. The choice is yours to provide the player with a fun and challenging experience.
```
MSilva95, line 121: as we ask this in the audit:
" Does the level include elements like moving platforms or skill-based platforming challenges?"

we should make it clear in this section of the readme that the student needs to implement some of these features

MSilva95 (changes requested):
Overall, it's very good!
I just pointed 2 things

subjects/gaming/2.5d-adventure/README.md
EldritchGriffin: i wanted to keep level design to be a creative process for the student hence not telling them exactly what elements to implement, should i rephrase it in the audit to make it so that they can implement whatever mechanics they want ? because thats what i was initially going for, and the moving platforms and skill based platforming were just examples.

MSilva95: yes, you can do that as well.
I understood but then they would probably complain that we didn't ask specifically for that on the readme

subjects/gaming/2.5d-adventure/audit/README.md
EldritchGriffin: okay, i thought it was clear that picking up a collectible -whatever its source-  should update the HUD.
but sure i will explicitly ask them to do that.

MSilva95: that is clear, the part of the `collectible` being spawned from killing is the part not mentioned there, it is mentioned in other section, you can also move the question down to that section, it is up to you, let me know what you think its better

EldritchGriffin: i mentioned it in the enemy section as its related to it.
i think i will add the following under where im asking them to spawn it in the subject:

> Tip: Make sure the instance of `collectible` spawned from killing an enemy also updates the HUD.

subjects/gaming/2.5d-adventure/README.md
EldritchGriffin: `Does the level include elements like moving platforms or skill-based platforming challenges?`

maybe this could be rephrased to :

`Does the level include mechanics chosen by the student?`

this will make it mandatory to implement some mechanic but not exactly specifying one.

or should i find another way to say it in the subject too?

what do you think?

MSilva95 (approved):
LGTM!

PR #2739, opened 2024-09-26
CON-3183 Fixing instructions in `AI.GO/only-if`

Oumaimafisaoui:
### Why?

The instructions were not clarifying enough what will be tested.This goes with tests following the old exercise : special promo.

### Solution Overview

Added clear instructions that are compatible with the tests.

### DIFF
--- a/subjects/AI.GO/only-if/README.md
+++ b/subjects/AI.GO/only-if/README.md
@@ -113,13 +113,19 @@ Your Robot must always seek the truth.
 
 #### Task 2:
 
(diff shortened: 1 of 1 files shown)


MSilva95 (approved):
LGTM!

Frenchris (approved):
I changed the grammar and a bit of the formatting more towards the end of the subject. 
If we keep the variable ticket and ask the student to declare it then it should be removed from the tests where it is pre declared. I ll see to it with @MSilva95

PR #2744, opened 2024-09-30
[EXTERNAL]  Clarify the need to implement the create_date function for birth date conversion

amali01:
### Why?

> The `create_date` function is necessary to properly convert a string (e.g., "2015-09-05") into a `NaiveDate` format for the `Form` struct's `birth` field. Without this function, the date field cannot be processed correctly, and the form validation will not work as expected. 

This update ensures that all the requirements of this question are clear, avoiding confusion and time waste.

### DIFF
--- a/subjects/error_types/README.md
+++ b/subjects/error_types/README.md
@@ -5,6 +5,7 @@
 For this exercise, you will have to implement an **error type** for a form validator. This must validate the password and the first name.
 
 The first name must not be empty and the password must have **at least 8 characters**, and a combination of **alphabetic**, **numeric** and **none-alphanumeric** (`<`, `&`,  `/` ...).
+Additionally, you need to create a helper function create_date that converts a string like "2015-09-05" into a NaiveDate format.
 
 Examples:
 - `"asDd123=%"`: **good**.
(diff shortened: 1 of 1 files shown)


subjects/error_types/README.md
```diff
@@ -5,6 +5,7 @@
 For this exercise, you will have to implement an **error type** for a form validator. This must validate the password and the first name.
 
 The first name must not be empty and the password must have **at least 8 characters**, and a combination of **alphabetic**, **numeric** and **none-alphanumeric** (`<`, `&`,  `/` ...).
+Additionally, you need to create a helper function create_date that converts a string like "2015-09-05" into a NaiveDate format.
```
zanninso, line 8: The question for this exercise to build the `error type` form validator, the data will be given as input in the right format.
they can use `NaiveDate::parse_from_str(date_str, "%Y-%m-%d")` if needed for test so no need to create a function to converts the date

subjects/error_types/README.md
```diff
@@ -15,7 +16,7 @@ Examples:
 Create a structure named `Form` that will have the following fields:
 - `first_name`: `String`
 - `last_name`: `String`
-- `birth`: `NaiveDate` that will convert a string `"2015-09-05"` to a date of that format.
+- `birth`: `NaiveDate` (converted from a string using the create_date function).
```
zanninso, line 19: The description here have no meaning since we have a data type `NaiveDate`.
Please remove the description from here then i can approve your PR

zanninso (changes requested):

PR #2765, opened 2024-10-09
CON-3142 Refining firing-range project

EldritchGriffin:
Recreated the project's subject and audit with higher quality.


subjects/gaming/firing-range/README.md
```diff
@@ -1,39 +1,125 @@
-## FiringRange
+## Firing-range
```
zanninso, line 1: i see that different formats is used in the gaming project and the one that u chosed should be apply to all the other project.

subjects/gaming/firing-range/README.md
```diff
@@ -1,39 +1,125 @@
...
+- Add multiple weapons (e.g., a shotgun, sniper rifle) with unique shooting and reload mechanics.
+- Implement more advanced target AI, such as random movement patterns or different difficulty levels.
+- Add different hit zones for targets, with headshots awarding more accuracy points.
+- Design a timed challenge mode where the player must hit all targets within a set time frame.
 
-#### Bonus
+### Submission
 
-Here are some ideas for improving the game:
+- You must upload a zip file of your game build in your repository.
```
zanninso, line 115: Is it not better to have just the code then rebuild the project during the audit ?

zanninso (approved):
I left two comments reply to it then you can merge it.

subjects/gaming/firing-range/README.md
EldritchGriffin: an unreal engine project is significantly larger than the build. the build only packages the necessary parts of the engine that were used in your game, and a build process can take quite long.

subjects/gaming/firing-range/README.md
EldritchGriffin: yes, i will be using kebab case in the project repos, but the names inside the subject file sometimes look weird when applying kebab case thats why i go for normal text, like 2.5D Adventure (2-5-d-adventure in kebab case)

subjects/gaming/firing-range/README.md
zanninso: Okay thank for the infos 🙏

subjects/gaming/firing-range/README.md
zanninso: Just make sure to keep it consistent over the branch 👍

PR #2766, opened 2024-10-09
Make the sentence more inclusive in time

sadiqui:
Using "ever" broadens the scope of the question, making it more inclusive of all past experiences, while the previous phrasing implies only a more recent or continuous reflection. The modification emphasizes the idea of whether a person have ever thought about it at all.

### DIFF
--- a/subjects/its-a-match/README.md
+++ b/subjects/its-a-match/README.md
@@ -2,7 +2,7 @@
 
 ### Instructions
 
-Have you been pondering over the etymology of `grep`?
+Have you ever been pondering over the etymology of `grep`?
 
 Create 4 regular expression variables:
 


HarryVasanth (changes requested):
The construction "ever been pondering" is not typically used and sounds awkward. The continuous form with "been pondering" doesn't pair well with "ever" in this context. "Ever" is usually used with the present perfect simple tense "have pondered" rather than the present perfect continuous "have been pondering".

sadiqui:
> Using "ever" broadens the scope of the question, making it more inclusive of all past experiences, while the previous phrasing implies **only** a more recent or continuous reflection, focusing on a continuous action just before now. The modification emphasizes the idea of whether a person have **ever** thought about it at all.

To keep the inclusivity of time while maintaining natural phrasing, it could be better as "**Have you ever pondered the etymology of grep?**"

HarryVasanth (approved):

PR #2834, opened 2025-01-01
fix(rust-roman_numbers): remove conflicting traits from expected data structure

heshamalmosawi:


pedrodesu (approved):

pedrodesu:
Looks good to me! Thank you very much for spotting this mistake and submitting the PR. I should also warn you the team is working on refining and updating the Rust content comprehensively at the moment, so this type of mistake should be fixed promptly, as with other improvements that will be curated.

PR #2862, opened 2025-01-30
fix(rust-quest03): re-added removed subjects

pedrodesu:
These subjects are deemed necessary for this branch for now so things don't break. From now on changes will be applied on a new branch.


MSilva95 (approved):
LGTM!

PR #2882, opened 2025-02-19
fix: remove check for function length in more-or-less

xySaad:
remove the test for add.length and sub.length because it's unnecessary.
example of code that wouldn't pass the test:

```js
const more=(arg)=>arg+1
const less=(arg)=>arg-1
const add=(...args)=>args[0]+args[1]
const sub=(...args)=>args[0]-args[1]
```

### DIFF
--- a/subjects/cybersecurity/hole-in-bin/README.md
+++ b/subjects/cybersecurity/hole-in-bin/README.md
@@ -42,7 +42,7 @@ Download the provided VM image and set it up in VirtualBox or UTM:
 
 - **SHA1 Checksums**:
 
-  - SHA1 for `hole-in-bin.ova`: 7db09b7a8fdfe25c286561dfa7ca5b50718bd60c
+  - SHA1 for `hole-in-bin.ova`: 00fda7d71361240d4d32499eb7fc5b156bbd53fc
   - SHA1 for `hole-in-bin.utm.zip`: fc93533b2054d10d03b09d53c223e57bf7ac7b62
 
 This VM contains all the binaries you will need to exploit.


sadiqui:
Despite that the code indeed adds the two first arguments, it isn't correct because it doesn't strictly enforce the requirement that `add` and `sub` must take exactly two arguments.

The [subject](https://github.com/01-edu/public/tree/master/subjects/more-or-less) states:
- `add` that takes 2 arguments and adds them together.
- `sub` that takes 2 arguments and subtracts the second argument from the first.

By using the rest parameter (`...args`), the function can accept any number of arguments, which contradicts the exercise's explicit instruction that **it should take two**. This means that calling `add(1, 2, 3)` or `sub(5)` won't result in an error but will instead ignore extra arguments or produce `NaN` when arguments are missing. A correct implementation should explicitly define two parameters (`(a, b) => a + b`), ensuring that the function **strictly** adheres to the requirement.

xySaad:
> @xySaad, despite that your code indeed adds the two first arguments, it isn't correct because it doesn't strictly enforce the requirement that `add` and `sub` must take exactly two arguments.
> 
> The [subject](https://github.com/01-edu/public/tree/master/subjects/more-or-less) states:
> 
> * `add` that takes 2 arguments and adds them together.
> * `sub` that takes 2 arguments and subtracts the second argument from the first.
> 
> By using the rest parameter (`...args`), the function can accept any number of arguments, which contradicts the exercise's explicit instruction that **it should take two**. This means that calling `add(1, 2, 3)` or `sub(5)` (in your case) won't result in an error but will instead ignore extra arguments or produce `NaN` when arguments are missing. A correct implementation should explicitly define two parameters (`(a, b) => a + b`), ensuring that the function **strictly** adheres to the requirement.

Even if the function takes only two arguments nothing is enforced, I can still pass any number of argument with no errors.
So it doesn't make sense to enforce the argument length in fucntion declaration.
As you know JavaScript follows permissive arity means i can pass extra arguments to rhe function with no error unless explicitly handled in the function, for example throw an error if arguments.length > Function.prototype.length

sadiqui:
@xySaad, in a JavaScript exercise, following the instructions as given is key to good practice. The task explicitly asks for a function that "takes 2 arguments," and defining it as `add(a, b)` correctly reflects that requirement. While JavaScript allows extra arguments, using `(...args)` suggests **unnecessary flexibility**, making the function **less clear**. Good practice in exercises means writing code that matches the specification while maintaining readability and intent. Additionally, `add(a, b)` properly sets `add.length` to `2`, preserving useful metadata, whereas `(...args)` **obscures the intended arity**.

PR #2892, opened 2025-02-27
DEV-6587: Cybersecurity hole in bin SHA1 don't match

zanninso:
DEV-6587

changing the hash of the vm to make it match the vm.


EldritchGriffin (approved):
SHA matches now, LGTM

PR #2900, opened 2025-03-06
[EXTERNAL] `make-your-game` Direct permanent link to image

sadiqui:
This pull request updates the image reference for the "59fps meme" by replacing the local file link (`59fps meme`) with a direct permanent URL (`59fps meme`). This change ensures that the image is now hosted online and accessible through a permanent link, making it easier to share and reference without relying on local resources. **Note:** with the current method, the image may not be displayed in some campuses or environments where the `resources/` directory is not imported into the students platform.


sadiqui:
> The recent issue of missing images is more likely related to the platform's development line rather than a content issue. I suggest holding off on merging this PR for a while.

@HarryVasanth @MSilva95
Just to clarify the issue: the images are not appearing in **any** of the subjects, not just one. Here's an example:

Screenshot from 2025-05-09 12-43-25

Clicking the image link redirects to the image path on Gitea:

Screenshot from 2025-05-09 12-48-11

So this issue should be fixed another way, there's nothing to commit or merge from here, **I'd suggest closing this PR**.

But I’d like to know: is the image issue only happening in the campus here, or is it the same in other campuses too?

HarryVasanth:
> But I’d like to know: is the image issue only happening in the campus here, or is it the same in other campuses too?

Hello @sadiqui, thank you for flagging this issue. It has been identified in other learning centres as well and has been resolved in our upcoming platform release.

We appreciate your contribution and time.

PR #2909, opened 2025-03-24
CON-3319 Fix `quest-05`

pedrodesu:


pedrodesu:
False positive. The links are indeed valid, they aren't broken. This action should be corrected.

MSilva95 (changes requested):
Apply all changes requested in module here as well.

MSilva95 (approved):

PR #2919, opened 2025-04-08
[EXTERNAL] docs(`typing-in-progress`): delay on pause

sadiqui:
This update ensures that the animation is not displayed when the user stops typing or finishes the conversation.

Optionally, a short delay can be added to handle natural pauses and improve UX, similar to how apps like Discord delay the disappearance of the "user is typing" indicator to avoid abrupt changes during brief pauses in typing.

This delay is not enforced, allowing for flexibility.


MSilva95 (approved):
LGTM!

PR #2931, opened 2025-04-16
CON-3320 Fix `quest-06`

pedrodesu:
[CON-3320](https://01talent.atlassian.net/browse/CON-3320)


MSilva95 (changes requested):
There is a conflict in the PR, I will continue the review, but in the meantime it needs to be fixed

pedrodesu:
Conflict resolved.

PR #2949, opened 2025-05-07
[EXTERNAL] 0-shell: bringing back the low-level system calls bonus for consistency

heshamalmosawi:
### Why?

This bonus was available around two weeks ago prior to the last doc update to the project requirement file. It was removed from the subject (perhaps by accident), however it is in the audit questions as a bonus, making an inconsistency between the readme and the audit questions. Adding this back will allow students to be aware of the bonus question prior to submission.

### Solution Overview

I just was copied the bonus point back from two versions ago from the file.

### DIFF
--- a/subjects/0-shell/README.md
+++ b/subjects/0-shell/README.md
@@ -57,6 +57,9 @@ Additional constraints:
 Implementing any of the following will be considered bonus:
 
 - Handle `Ctrl+C` (SIGINT) without crashing the shell
+- Implement the commands exclusively using `low-level system calls` avoiding built-in functions or libraries that abstract file operations.
+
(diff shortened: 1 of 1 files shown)


pedrodesu:
Hello @heshamalmosawi!

Thank you for your suggestion, as always it's very much appreciated. Indeed we removed the "low-level system calls" part from the project deliberately, as we're shifting the project's language to Rust and we felt like this bonus no longer felt like a part of the scope of the project or its expected time frame. There was also a certain ambiguity regarding exactly what should and should not be implemented using system calls, and where.

All of this composes the rationale of removing this from the bonus. Naturally the student can still do it, but it's not what the bonus expects, and there's still a lot of more relevant and better defined tasks on the bonus part of the subject.

Regarding the inconsistency problem you pointed out, it seems like something we might have overlooked - As such this part of the bonus won't be added back, but we will solve this inconsistency you pointed out between the subject and the audit shortly.

Thank you for helping us improve our content and please let me know if I missed something or something was left improperly explained.

TheJaberi (commented):
Cool idea!!

PR #2961, opened 2025-05-29
[EXTERNAL] docs(`mini-framework`): fix subject and audit file

sadiqui:
Fix minor documentation issues: standardized the "Routing System" header format, corrected a typo ("auditers" → "auditors"), and updated the TodoMVC example link to a valid one.

### DIFF
--- a/subjects/mini-framework/README.md
+++ b/subjects/mini-framework/README.md
@@ -9,7 +9,7 @@ Be aware that a framework is different from a library. When you call a method fr
 Your framework should implement:
 
 - Abstracting the DOM
-  Routing System
+- Routing System
 - State Management
 - Event Handling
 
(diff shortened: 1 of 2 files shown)


Muaytie666 (dismissed):

sadiqui:
@HarryVasanth @Muaytie666 Does this require a new review?

The merge is necessary because the current state of the README creates confusion about the **minimum requirements** for the **Mini-Framework** project, due to a markup syntax error.

The project expects the framework to handle **four distinct features**:

* Abstracting the DOM
* Routing System
* State Management
* Event Handling

However, because of a missing dash before "Routing System", the list currently renders as (refer to the attached image):

* Abstracting the DOM  Routing System
* State Management
* Event Handling

This makes it appear as if "Abstracting the DOM Routing System" is one item, which is misleading, as there's **no such thing** conceptually. "Abstracting the DOM" and "Routing System" are **two separate** and essential requirements.

This error may lead some teams to mistakenly implement only three of the four required features.

image

zanninso (approved):
LGTM!

PR #2996, opened 2025-07-08
[EXTERNAL] docs(`printifnot`): less ambiguous instructions

sadiqui:
Replace with clearer, unambiguous instructions that explicitly state all requirements and return cases.

- **Note:** This change does not affect [test compatibility](https://github.com/01-edu/go-tests/blob/master/tests/printifnot_test/main.go) or [expected solutions](https://github.com/01-edu/go-tests/blob/master/solutions/printifnot.go), as the requirements themselves remain unchanged.

### DIFF
--- a/subjects/box_recursion/README.md
+++ b/subjects/box_recursion/README.md
@@ -30,11 +30,11 @@ pub struct WorkEnvironment {
     pub grade: Link,
 }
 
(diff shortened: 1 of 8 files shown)


EldritchGriffin (approved):
LGTM

PR #3010, opened 2025-07-28
[EXTERNAL] docs(modify-letter): Changed to the correct output

ICEBergJQ:
output section the first out put should be "John" instead of "Joe"

> Before starting, please choose the relevant pull request **Labels**, **Reviewers**, and **Assignees** </br>
> Here are a couple of example tiles of a good pull request:
> - `CON-1234 Title of The Feature or Fix`
> - `[EXTERNAL] Title of The Feature or Fix`

CON-#### | [EXTERNAL]

### Why?

### Solution Overview

### Implementation Details

### Build Images

>- `💻 Sh` 
>- `🚀 JS`
>- `🧩 DOM`

### DIFF
--- a/subjects/modify_letter/README.md
+++ b/subjects/modify_letter/README.md
@@ -40,7 +40,7 @@ And its output
 
 ```console
 $ cargo run
-Joe is sleeping
+John is sleeping
 Jim is swimming
 bye byE
 $


pedrodesu:
Hello @ICEBergJQ! Indeed this is an issue we've overlooked. Thank you for reaching out! I will approve the PR.

pedrodesu (approved):
Looks good.

PR #3013, opened 2025-08-01
[EXTERNAL] docs(`drawing`): fix implementation example for rectangle

sadiqui:
### Fix Rectangle Implementation Example

This PR updates the example usage of the `Rectangle` shape to correctly demonstrate a **rectangle**, rather than a **square**, as the original coordinates resulted in equal width and height.

### Changes:

* Replaced:

  ```rust
  let rectangle = gs::Rectangle::new(&gs::Point::new(150, 150), &gs::Point::new(50, 50));
  ```

  with:

  ```rust
  let rectangle = gs::Rectangle::new(&gs::Point::new(150, 300), &gs::Point::new(50, 60));
  ```

### Note:

This corrected example is borrowed from the [rust/tests](https://github.com/01-edu/rust-tests/blob/master/solutions/drawing/src/main.rs) repository.


pedrodesu:
While technically squares classify as rectangles, I agree changing to a non-square example, the same example to the one provided in the tests here makes sense. If you can, please run `prettier -w` on the file and push the change, as it's not Prettier compliant as of right now. With that I can merge without conflicts. Thank you once again!

PR #3016, opened 2025-08-02
fixed a wrong data type given in the instructions.

madaghaxx:
The instruction were requesting a HashMap with a tuple of strings as key and a value of a callback function but the expected function used a HashMap with string as key and callback as value.
The given instructions where misleading the users.

## The instructions:
- A second structure named `FlagsHandler` will be given which just has one element: `flags: HashMap<(String, String), Callback>`. You'll also need to implement the following associated **functions**:

## The given expected function:
```rust
pub struct FlagsHandler {
    pub flags: HashMap<String, Callback>,
}
```
So the fixed version of the instructions is :
- A second structure named `FlagsHandler` will be given which just has one element: `flags: HashMap<String, Callback>`. You'll also need to implement the following associated **functions**:

### DIFF
--- a/subjects/banner/README.md
+++ b/subjects/banner/README.md
@@ -22,7 +22,7 @@ println!("short hand: {}, long hand: {}, description: {}", d.short_hand, d.long_
 
 An associated **type** called `Callback` will also be provided. It should represent a function pointer which is going to be used in the structure and functions below. This function will represent the callback for the flag associated to it.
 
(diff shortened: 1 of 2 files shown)


pedrodesu:
Hello @madaghaxx! Thank you for the PR. Everything looks good, and so we will merge this. The action failed but it's a false positive. We will report this too.

Let me know if you have anything else to add.

PR #3019, opened 2025-08-04
CON-3607 Fix retrieve-transaction-in-out subject inconsistency name between test and subject

zanninso:
CON-3607


EldritchGriffin (approved):
LGTM

PR #3023, opened 2025-08-10
[EXTERNAL] Update README.md

kinoz01:
based on tests, for example:
assert_eq!(0, prev_prime(2));
assert_eq!(2, prev_prime(3));
assert_eq!(3, prev_prime(5));

here it should be "strictly less than", not "or equal to"

### DIFF
--- a/subjects/previousprime/README.md
+++ b/subjects/previousprime/README.md
@@ -2,7 +2,7 @@
 
 ### Instructions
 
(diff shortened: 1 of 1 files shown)


pedrodesu:
This is correct, however this issue was already fixed. Closed in favor of https://github.com/01-edu/public/pull/3043. Thank you @kinoz01!

PR #3047, opened 2025-08-26
fix(numpy): clarify CSV header exclusion in instructions

vpollo11:
#### What changed

The sentence:

> "Load the data using genfromtxt, specifying the delimiter as ';', and optimize the numpy array size by reducing the data types. Use np.float32 and verify that the resulting numpy array weighs 76800 bytes."

was unclear about whether the CSV header should be skipped or not.

It now says:

> "Load the data using genfromtxt, specifying the delimiter as ';' with excluding the headers, and optimize the numpy array size by reducing the data types. Use np.float32 and verify that the resulting numpy array weighs 76800 bytes."

#### Why

Later instructions refer to specific row numbers (like the 2nd, 7th, and 12th rows). Without stating whether the header counts as a row, the meaning is ambiguous — it affects the indexing. This fix makes it clear that the header should be excluded.

#### No code changes

This is a documentation fix only.

### DIFF
--- a/subjects/ai/numpy/README.md
+++ b/subjects/ai/numpy/README.md
@@ -296,7 +296,7 @@ The goal of this exercise is to perform fundamental data analysis on real data u
 
 The dataset chosen for this task was the [red wine dataset](./data/winequality-red.csv). You can find more info [HERE](./data/)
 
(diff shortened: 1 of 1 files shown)


EldritchGriffin (approved):
LGTM 🚀

PR #3066, opened 2025-09-22
[EXTERNAL] fix(pandas): replace broken "Ultimate Pandas Resource" link with "Awesome Pandas"

vpollo11:
Thanks for catching this!  
The "Ultimate Pandas Resource" link was returning a 403 error.  
I’ve updated it to point to the actively maintained [Awesome Pandas](https://github.com/tommyod/awesome-pandas) resource instead.


Oumaimafisaoui (approved):
LGTM

PR #3112, opened 2025-11-19
[EXTERNAL] Refactor constructor names in Decorator classes

Mellagui:
Renamed constructors in Decorator classes to follow consistent naming patterns, making the codebase more uniform and easier to understand.


MSilva95:
Hello @Mellagui 
We will update it in the correct repo since this one is no longer maintained.
thank you for the feedback!

PR #3113, opened 2025-11-22
[EXTERNAL] Removes some duplicate files

motonai:
[EXTERNAL]

I noticed in `groupie-tracker` (but also in other subjects) that there are duplicate files inside the project's directories. Apparently, there is this pattern:
```
subjects/<project>/<subproject>/{audit.md,audit/README.md}
```
The only think changing between them is the link to the `good-practices` document.

### Why?

They seem to serve no purpose. As far as other projects' structures suggest, the `README.md` file under the `audit` directory of a `subproject` is the one that is valid.

### Solution Overview

Removal of the duplicates for the `groupie-tracker`'s subprojects.

### Implementation Details

In the cases included in the PR, I choose to remove only the ones for `groupie-tracker` project.

I choose to not remove other duplicates, since there are some cases that the differences exceed the link to the `good-practices` document. For example, in the subproject `0-shell/job-control`, one suggests to run `ls` while the other suggests `tar`. While the convention seem to be straight-forward, for those that differ more than the link, I wanted to check with you first and I can append the current PR or revisit on a new one.

I found those with the following:
```
find . -type f | grep audit.md | while read line; do echo $line | cut -d '/' -f 3,4; echo "==================="; diff $line ${line/.md//README.md}; echo; done
```


motonai:
I close this since the repo is unmaintained

PR #3133, opened 2026-01-19
fix(rust-piscine): add missing std::rc::Rc import

aymeenOvski:


MSilva95:
Hello @aymeenOvski 
thank you for the feedback, i will fix it in the correct repo since this one is no longer maintained.

PR #3135, opened 2026-01-20
[EXTERNAL] fix: clarify file-details subject hints misleading comment  and correct awk examples qouting

e-aub:
### Why?  
The original `file-details/README.md` exercise instructions contained misleading comments about `awk` field usage and included incorrect quotes in examples. This could confuse learners and cause syntax errors when following the hints.  

### Solution Overview  
- Corrected the comments describing `awk` fields to reflect what is actually printed.  
- Replaced smart quotes with straight quotes in all `awk` examples.  
- Reorganized examples for better readability and clarity for beginners.  

### Implementation Details  
- Updated all `awk` example commands to use proper shell-compatible quotes. 
- Clarified which columns are selected in each `awk` example, avoiding misleading “all fields” comments.


MSilva95:
Hello @e-aub 
This is an older version of the exercise, i did find a couple of smart-quotes that i will remove in the correct repo since this repo is no longer maintained. 
thank you for the feedback!

Issue #3163, opened 2026-04-14
[BUG]


EnTechGR:
**Describe the bug**
Fact Correction: Domain Names vs. IP Addresses

**Users**
All

**Severity**
❗️minor

**Type**
🗂️ documentation

**To Reproduce**

1. Go to 'https://github.com/01-edu/public/tree/master/subjects/login'
2. Scroll down to line 45 of README
4. See error:
   "- A domain name can have several IP addresses, but an IP address can only identify one domain name"

**Workarounds**
Actually, one IP address can identify many domain names. This is how "Virtual Hosting" works. A single server (one IP) can host hundreds of different websites (e.g., site-a.com and site-b.com might both point to 192.168.1.1). The server uses the HTTP header to figure out which site the user is actually looking for.

**Expected behavior**
-

**Attachments**
-

**Additional context**
-

motonai:
As you say (and so says the file): A domain name can have several IP addresses!!

You support that this part is wrong: "but an IP address can only identify one domain name", which is not.

Your explanation still confirms the first part but doesn't prove the second one. For the second part, you have to refer to DNS PTR records.

My proposal is to close the issue; it's not a bug.

References:
- https://www.nslookup.io/learning/dns-record-types/ptr/
- https://datatracker.ietf.org/doc/html/rfc1034
- https://datatracker.ietf.org/doc/html/rfc1035

PR #3164, opened 2026-04-14
Fix issue #3163: corrected the IP-domain rlation statement

EnTechGR:
CON-3163 | Corrected IP-Domain mapping documentation
====================================================

### Why?

The existing documentation contained a logical inaccuracy regarding the relationship between domain names and IP addresses. It stated that an IP address can only identify one domain name, which ignores the industry-standard practice of **Virtual Hosting**. This change ensures the technical documentation correctly reflects how modern web servers and DNS function.

### Solution Overview

I have updated the descriptive text to clarify that the relationship between domain names and IP addresses is **many-to-many**.

The corrected statement now explains that:

1.  **One Domain** can resolve to **Multiple IPs** (for load balancing).
    
2.  **One IP** can host **Multiple Domains** (via Virtual Hosting/SNI).
    

### Implementation Details

The revision replaces the restrictive "one-to-one" implication with a more accurate "many-to-many" explanation.

*   **Approach Chosen:** Directly addressing the "Virtual Hosting" concept. This was chosen over a simple deletion because explaining _why_ the numbers match the names provides better educational value for the project.
    
*   **Alternatives Considered:** I considered adding a deep dive into SNI (Server Name Indication), but decided against it to keep the "Just Numbers" section concise and focused on the core concept of identifiers.

### DIFF
--- a/subjects/login/README.md
+++ b/subjects/login/README.md
@@ -42,7 +42,7 @@ Behind every name in a computer system there is a number (ID, index, address, et
 
 Names exist because they are human readable, but behind the scenes they are converted into numbers, unique in their namespace :
 
-- A domain name can have several IP addresses, but an IP address can only identify one domain name
+- A domain name can map to several IP addresses (for load balancing), and a single IP address can host many different domain names (via Virtual Hosting).
 - Several processes may have the same name, but a PID identifies a single process
 
 Find the commands to get :


motonai:
The "industry-standard of virtual hosting" has nothing to do with the fact that an IP can *point* only to one domain. For more info, read the reply at #3163 and its references.

Silas-Labs:
This is a good insight. Traditionally, before the advent of VPS and load balancers and such, the relationship was always a 1-1

motonai:
It's about the way DNS works. While you could always assign many domains to the same IP, the IP can have one PTR record that points to a domain. Hence:

- Many domain names -> 1 IP
- 1 IP -> 1 Domain name

You can see this happening when doing SMTP server setup for multiple domains. Because of PTR, the allowed domain is the one found in PTR record. Adding a TXT record for setting SPF is needed to grand permission for extra domains.

## For example
- you have your IP pointing at `example.com` (PTR record),
```
2.0.1.192.in-addr.arpa. PTR example.com.
```
- you also have `example.com` point to your IP (A/AAAA records),
```
example.com A 192.1.0.2
```
### Then you want to support mail for `newexample.com`
- you set up `newexample.com` point to your IP (A/AAAA records),
```
newexample.com A 192.1.0.2
```
- you add `example.com` to your `newexample.com` SPF (TXT record).
```
newexample.com TXT v=spf1 include:example.com ~all
```
## TLDR
The relation 1IP-has-1domain is still in place. That is different from many-domains-have-one-IP. It's not bidirectional relationship. The first is also called reverse DNS, while the latter is forward one.

MSilva95:
Hello @EnTechGR !
This branch was updated and as consequence the project in question changed as well, this issue is no longer present in the new version, thank you for your feedback!

EnTechGR:
Hey! Thanks for the feedback, I really appreciate it.

PR #3187, opened 2026-07-11
docs(ion-out): clarify that words must end in tion

skamprogiannis:
**Title:**

```
[EXTERNAL] Clarify ion-out README wording
```

### Why?

The current README is ambiguous. It says to return every word containing `ion` following a `t`, with the `ion` part removed. Read literally, this implies that words such as:

```text
mentions -> ments
```
should be included. However, the test suite only accepts words where `tion` occurs at the end of the word. 

### Solution Overview

Clarify that only words ending in `tion` are expected. The revised wording makes the documented requirements match the tests and makes the intended use of a lookahead clearer.

### DIFF
--- a/subjects/ion-out/README.md
+++ b/subjects/ion-out/README.md
@@ -2,7 +2,7 @@
 
 ### Instructions
 
-Create a function named `ionOut`, that receives a `string` and returns an array with every word containing `'ion'` following a `'t'`. The words should be returned without the `'ion'` part.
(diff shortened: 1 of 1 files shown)


Silas-Labs (commented):
Your rephrase matches exactly the original in semantics

MSilva95:
hello @skamprogiannis!
This is an older version of the exercise, this repo is no longer maintained so it contains older versions of some exercises and projects. If you check the exercise out, it will have a new readme. Thank you for your feedback!

PR #3189, opened 2026-07-30
Update README.md

kill-ux:
> Before starting, please choose the relevant pull request **Labels**, **Reviewers**, and **Assignees** </br>
> Here are a couple of example tiles of a good pull request:
> - `CON-1234 Title of The Feature or Fix`
> - `[EXTERNAL] Title of The Feature or Fix`

CON-#### | [EXTERNAL]

### Why?

### Solution Overview

### Implementation Details

### Build Images

>- `💻 Sh` 
>- `🚀 JS`
>- `🧩 DOM`

### DIFF
--- a/subjects/mobile-dev/map-markers/audit/README.md
+++ b/subjects/mobile-dev/map-markers/audit/README.md
@@ -34,4 +34,4 @@
 
 ###### Does the map navigates to the tapped place?
 
-###### Does the app work both on android and ios?
+###### Does the app work both on android or ios?


Silas-Labs (commented):
The test is supposed to confirm the working on both devices and not on any of them...strictly both.

kill-ux:
> The test is supposed to confirm the working on both devices and not on any of them...strictly both.

But in our building we get an iphone to test apps in it but for testing it we need MacOs and in our compus we have just a Linux so that impossible i see that we just add the code of ios but we don't test it.

MSilva95:
hello @kill-ux!
I understand the issue you are having so i tried to create a solution for it by adding the following to the audit:
```
##### Cross-Platform

###### Does the app work on both Android and iOS? (Follow the instructions below)

1. Test the app on both Android and iOS when both platforms are available.
2. If the audit environment does not allow testing on one of the platforms, test the available platform and inspect the code and platform-specific configuration of the untested platform.
3. If the tested platform works correctly and the auditor verifies that the required implementation for the untested platform is present and correctly configured, consider this criterion fulfilled and validate the question.
4. Do not validate the question if the untested platform is missing the required implementation or configuration.
```

I will close this pr since this repo is no longer maintained and the new pr is in review. 
Thank you for the feedback

